PR #5, opened 2021-08-13
Quick fixes

ru-fu:
Two commits with original and update to make it easier to check the changes.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -1,0 +1,115 @@
+This guide provides the first steps to using Anbox Cloud. If you haven't installed Anbox Cloud, please refer to the [installation page](https://discourse.ubuntu.com/t/installation-quickstart/17744) first.
+
+## Access to AMS
+
+For all subsequent commands using the `amc` tool to work you need to access the `ams/0` machine on a regular Anbox Cloud deployment. You can do this by opening an SSH session with the `juju` command:
+
+```bash
+$ juju ssh ams/0
+```
+
+If you're running the Anbox Cloud Appliance instead, you can find the `amc` tool directly on the host and it's already setup to talk to the deployed AMS service.
+
+Alternatively you can also install the `amc` command on your local Ubuntu based development machine. See [AMS Access](https://discourse.ubuntu.com/t/managing-ams-access/17774) for more details.
+
+## Ensure Images are Available
+
+As a next step you can check that **AMS** has synchronized all images from the Canonical hosted
+image server. You can list all synchronized images with the following command:
+
+```bash
+$ amc image ls
++----------------------+-----------------------------+--------+----------+---------+
+| ID                   | NAME                        | STATUS | VERSIONS | DEFAULT |
++----------------------+-----------------------------+--------+----------+---------+
+| bpkam07d6hg5q3ihfglg | bionic:android7:amd64       | active | 1        | true    |
++----------------------+-----------------------------+--------+----------+---------+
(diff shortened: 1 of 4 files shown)


overview.md
```diff
@@ -0,0 +1,58 @@
+Anbox Cloud provides a rich software stack to enable you to run Android in the cloud for all kinds of different uses cases. This page intends to give you an overview over the available variants and internal components which make Anbox Cloud.
+
+## Variants
+
+Anbox Cloud exists in two different variants which serve different purposes:
+
+**Anbox Cloud Appliance**
+
+The Anbox Cloud Appliance is a self-contained, single-machine deployment variant of Anbox Cloud. It uses a sensible standard configuration to hide the complexity and flexibility of Juju. In this way, the Anbox Cloud Appliance allows simple and fast development and prototyping.
```
ru-fu, line 9: Updated this description. Are we okay with this wording?

overview.md
```diff
@@ -0,0 +1,58 @@
...
+Anbox Cloud takes care of all management aspects and provides both a fully functional implementation and an integration model to support existing infrastructure and service implementations.
+
+The following figure gives an overview over the different components and their responsibility within the core stack of Anbox Cloud.
+
+![anbox-core-overview|690x398](upload://x0budFKybsbYLhiplNhF4R91Ght.png)
+
+At the heart of Anbox Cloud sits the **Anbox Management Service (AMS)**. **AMS** has the job to handle all aspects of the application and container lifecycle (including application and image updates) while ensuring high density, performance and fast container startup times.
+A developer or system administrator will manage **AMS** through the **command line interface (AMC)** or through custom-built tools interacting with the [**AMS REST API**](https://discourse.ubuntu.com/t/ams-rest-api-reference/17801).
+
+For example, a simple Android application testing service would provide a user-facing interface dealing with things like authentication and user management, and would communicate with the REST API to add applications or start and stop containers when a user asks to.
```
ru-fu, line 32: This was "...communicate with the REST API to add applications, start and containers when a user asks to.". Is the update correct?

getting-started.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Ensure Images are Available
+
+As a next step you can check that **AMS** has synchronised all images from the Canonical hosted image server. You can list all synchronised images with the `amc image ls` command:
+
+```bash
++----------------------+-----------------------------+--------+----------+--------------+---------+
+| ID                   | NAME                        | STATUS | VERSIONS | ARCHITECTURE | DEFAULT |
++----------------------+-----------------------------+--------+----------+--------------+---------+
+| c4b4djkrorjohh948dfg | bionic:android11:arm64      | active | 1        | aarch64      | true    |
```
ru-fu, line 23: I updated this table to be more recent, but noticed afterwards I see only the Arm images. Does this make sense, or should the table show the amd ones?

overview.md
morphis: Sounds good to me. Do we maybe want to link to Juju here to provide more context if needed?

overview.md
morphis: Sounds good to me

getting-started.md
morphis: Showing one of both is fine. We could even mix arm and x86 if we want as all these variants are possible.

morphis:
@ru-fu Overall looks good to me. Feel free to merge and publish if you're ready.

PR #8, opened 2021-08-19
Quick fixes

ru-fu:
Check the "language fixes and updates" commit to see what actually changed.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -4,17 +4,15 @@ This guide provides the first steps to using Anbox Cloud. If you haven't install
 
 For all subsequent commands using the `amc` tool to work you need to access the `ams/0` machine on a regular Anbox Cloud deployment. You can do this by opening an SSH session with the `juju` command:
 
-```bash
-$ juju ssh ams/0
-```
+    juju ssh ams/0
 
 If you're running the Anbox Cloud Appliance instead, you can find the `amc` tool directly on the host and it's already set up to talk to the deployed AMS service.
 
 Alternatively you can also install the `amc` command on your local Ubuntu-based development machine. See [AMS Access](https://discourse.ubuntu.com/t/managing-ams-access/17774) for more details.
 
-## Ensure Images are Available
+## Ensure images are available
 
-As a next step you can check that **AMS** has synchronised all images from the Canonical hosted image server. You can list all synchronised images with the `amc image ls` command:
+As a next step you can check that AMS has synchronised all images from the Canonical hosted image server. You can list all synchronised images with the `amc image ls` command:
 
 ```bash
 +----------------------+-----------------------------+--------+----------+--------------+---------+
@@ -48,43 +46,47 @@ Once it is up and running you can get a shell inside the container with:
 See [Managing containers](https://discourse.ubuntu.com/t/managing-containers/17763) for more details.
 
 ## Create an application
-**AMS** provides functionality to manage Android applications for you. To let **AMS** manage your application, you need the APK of the application and a `manifest.yaml` which looks like this in its simplest form:
 
-```bash
-$ cat manifest.yaml
+You can create applications through the web-based dashboard or on the command line.
+
+Both alternatives require you to provide the name and the [instance type](https://discourse.ubuntu.com/t/instances-types-reference/17764) of the application. If you want to run a specific Android application, you need to provide its APK.
+
+> **Hint**: If your application requires the use of a GPU for rendering and video encoding, select an instance type with GPU support like `g2.3`. For other instance types, the container will use a GPU if available or software encoding otherwise.
+
+### Create an application through the web dashboard
+
+The web dashboard is available if you installed [Anbox Cloud with the streaming stack](https://anbox-cloud.io/docs/installation/installation-quickstart) or the [Anbox Cloud Appliance](https://anbox-cloud.io/docs/installation/installation-quickstart). See [Web Dashboard](https://anbox-cloud.io/docs/manage/web-dashboard) for more information about the dashboard.
+
+To create an application through the web dashboard, open `https://<your-machine-address>/applications` in your browser. Click **Add Application** and enter the required information. If you want to run a specific Android application, upload its APK.
+
+### Create an application on the command line
+
+To create and manage an application from the command line, use `amc`. 
+
+You must provide a `manifest.yaml` file for your application. In its simplest form, the manifest looks like this:
+
+```yaml
 name: com.my.game
 instance-type: a2.3
 ```
 
-> **Hint**: If you deployed Anbox Cloud with the streaming stack, use an instance type with GPU support like `g2.3`. Otherwise the container will not get access to a GPU for rendering and video encoding.
+It defines the name of the application and which instance type the application should use. The manifest can also contain more advanced configuration like [Addons](https://discourse.ubuntu.com/t/managing-addons/17759), permissions and others. You can find more details about the manifest format and the available instance types in the [Application Management](https://discourse.ubuntu.com/t/managing-applications/17760) and [Instance Types](https://discourse.ubuntu.com/t/instances-types-reference/17764) sections.
 
-The manifest basically defines the name of the application and which instance type the application should use as well as more advanced configuration like [Addons](https://discourse.ubuntu.com/t/managing-addons/17759), permissions and others . You can find more details about manifest format and the available instance types in the [Application Management](https://discourse.ubuntu.com/t/managing-applications/17760) and [Instance Types](https://discourse.ubuntu.com/t/instances-types-reference/17764) sections.
+To create an application, place the `manifest.yaml` file in a directory of your choice. If you want to run a specific Android application, place its APK in the same directory. Then run the following command, replacing */path/to/manifest/directory/* with the path to your directory:
 
-To create the application with `ams`, place the APK as `app.apk` in the same directory as the `manifest.yaml`, and run the following command:
+    amc application create /path/to/manifest/directory/
 
-    amc application create /path/to/directory/with/apk/and/manifest/
+The tool will now run through a bootstrap process for the application to allow for faster boot times of the application later on.
 
-**AMS** will now run through a bootstrap process for the application to allow for faster boot times of the application later on.
-
-You can monitor the progress of the application with:
+You can use the `amc application show <app name>` command to display the status of your application. To continually monitor the progress of the application, enter the following command:
 
     watch -n 1 amc application ls
 
-When the application is marked as `ready`, you just have to publish it and it's ready to be used:
-
-    amc application publish <app name> 0
-
-Now you can simply start a container for your new application by launching the application:
+When processing is done, you can start a container for your new application by launching the application:
 
     amc launch <app name>
 
-## Accessing the Web Dashboard
-
-The Streaming Stack ships with a user friendly web-based dashboard that can be used to create, manage and stream Android application to a web browser.
-
-You can read more on https://anbox-cloud.io/docs/manage/web-dashboard
-
-## Next Steps
+## Next steps
 * [Image Management](https://discourse.ubuntu.com/t/managing-images/17758)
 * [Container Management](https://discourse.ubuntu.com/t/managing-containers/17763)
 * [Application Management](https://discourse.ubuntu.com/t/managing-applications/17760)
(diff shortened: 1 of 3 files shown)


manage-images.md
```diff
@@ -0,0 +1,120 @@
+An image is the base for a container running in the Anbox Cloud. It contains all necessary components, like Anbox or the Android root file system. Each release of Anbox Cloud comes with an update image, which must be uploaded to the AMS service to make it available to be used inside any container.
+
+AMS can manage different types of images, which can be individually selected by applications. For that, each image must be named when it's initially uploaded to AMS. If the same image receives an update with the next release of Anbox Cloud, it does not need to be re-created but can simply be updated. This will cause all applications using the image to be updated and rebased on the new image version.
+
+Officially released images are available for each version of Anbox Cloud, either on the [Downloads](https://oem-share.canonical.com/partners/indore/share/docs/1.7/en/installation-downloads.html) page (for versions <= 1.7) or on an image server hosted by Canonical.
```
ru-fu, line 5: Is it correct that the Downloads page is only for versions up to 1.7?

manage-images.md
```diff
@@ -0,0 +1,120 @@
...
+| Name                        | Android Version | Ubuntu Version | Support Status     | Available since |
+|-----------------------------|-----------------|----------------|------------|---------------|
+| `bionic:android11:amd64`      | 11              | 18.04          | supported | 1.10 |
+| `bionic:android11:arm64`      | 11              | 18.04          | supported | 1.10 |
+| `bionic:android10:amd64`      | 10              | 18.04          | supported | 1.0
+| `bionic:android10:arm64`      | 10              | 18.04          | supported | 1.0
+| `bionic:android7:amd64`       | 7               | 18.04          | unsupported as of 1.10 | 1.0 |
+| `bionic:android7:arm64`       | 7               | 18.04          | unsupported as of 1.10 | 1.0 |
+
+The `bionic:android7:arm64` and `bionic:android7:amd64` images are not receiving Android security updates anymore as Google stopped supporting Android 7 in 2019. They are only provided for legacy reasons and have been dropped with Anbox Cloud 1.10 in April 2021.
```
ru-fu, line 22: Should these still be listed?

manage-images.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Delete an image
+
+Deleting an image will make it unavailable to Anbox Cloud. However, it will not affect any application based on the image directly, as the application keeps a copy of the image internally. As the image is not available to AMS anymore after deleting it, updating the underlying image of applications with a new version is not possible anymore.
+
+Deleting a specific image can be achieved with the following command, where `default` is the name of the image to delete:
+
+    amc image delete default
+
+Specific image versions can be deleted too, which is useful when all applications were migrated to a newer version and the old version is not needed anymore. The only requirement is that a single version of the image is available at all times.
```
ru-fu, line 111: The last sentence was "Only requirement is that an image as a single version available at all time." - not sure what that is supposed to mean. ;) Does my update capture it?

getting-started.md
```diff
@@ -48,43 +46,51 @@ Once it is up and running you can get a shell inside the container with:
 See [Managing containers](https://discourse.ubuntu.com/t/managing-containers/17763) for more details.
 
 ## Create an application
-**AMS** provides functionality to manage Android applications for you. To let **AMS** manage your application, you need the APK of the application and a `manifest.yaml` which looks like this in its simplest form:
+
+You can create applications through the web-based dashboard or on the command line with AMS.
+
+Both alternatives require you to provide the APK of the Android application, its name and its [instance type](https://discourse.ubuntu.com/t/instances-types-reference/17764).
+
+> **Hint**: If you deployed Anbox Cloud with the streaming stack, use an instance type with GPU support like `g2.3`. Otherwise the container will not get access to a GPU for rendering and video encoding.
```
aguadoenzo, line 54: This is misleading, you can deploy the streaming stack on a system without a GPU, which means that an application created with a `g*` instance-type won't run because of the lack of GPU.
Every instance-type uses a GPU if available and software encoder otherwise, `g*` instances **require** a GPU

getting-started.md
```diff
@@ -48,43 +46,51 @@ Once it is up and running you can get a shell inside the container with:
 See [Managing containers](https://discourse.ubuntu.com/t/managing-containers/17763) for more details.
 
 ## Create an application
-**AMS** provides functionality to manage Android applications for you. To let **AMS** manage your application, you need the APK of the application and a `manifest.yaml` which looks like this in its simplest form:
+
+You can create applications through the web-based dashboard or on the command line with AMS.
+
+Both alternatives require you to provide the APK of the Android application, its name and its [instance type](https://discourse.ubuntu.com/t/instances-types-reference/17764).
```
aguadoenzo, line 52: Technically, the APK is not needed. If not specified you are presented with the Android 10 launcher
In terms of tutorial however I'm not sure whether we should use an APK or not, depends on the level of simplicity we're aiming at

getting-started.md
```diff
@@ -48,43 +46,51 @@ Once it is up and running you can get a shell inside the container with:
...
+
+### Create an application through the web dashboard
+
+The web dashboard is available if you installed [Anbox Cloud with the streaming stack](https://anbox-cloud.io/docs/installation/installation-quickstart) or the [Anbox Cloud Appliance](https://anbox-cloud.io/docs/installation/installation-quickstart). See [Web Dashboard](https://anbox-cloud.io/docs/manage/web-dashboard) for more information about the dashboard.
+
+To create an application through the web dashboard, open `https://<your-machine-address>/applications` in your browser. Click **Add Application**, enter the required information and upload the APK of the Android application.
+
+### Create an application with AMS
+
+To create and manage an application from the command line, use AMS.
```
aguadoenzo, line 64: This could be confusing, the CLI is called `amc` not `ams`

getting-started.md
```diff
@@ -48,43 +46,51 @@ Once it is up and running you can get a shell inside the container with:
...
+
+To create an application through the web dashboard, open `https://<your-machine-address>/applications` in your browser. Click **Add Application**, enter the required information and upload the APK of the Android application.
+
+### Create an application with AMS
+
+To create and manage an application from the command line, use AMS. 
+
+AMS requires a `manifest.yaml` file for your application. In its simplest form, the manifest looks like this:
 
 ```bash
```
aguadoenzo, line 68:
```suggestion
```yaml
```

getting-started.md
```diff
@@ -48,43 +46,51 @@ Once it is up and running you can get a shell inside the container with:
...
 
-**AMS** will now run through a bootstrap process for the application to allow for faster boot times of the application later on.
+AMS will now run through a bootstrap process for the application to allow for faster boot times of the application later on.
 
 You can monitor the progress of the application with:
 
     watch -n 1 amc application ls
 
-When the application is marked as `ready`, you just have to publish it and it's ready to be used:
+When processing is done and the application is marked as `ready`, it is automatically published. Run the following command to check the publish state:
```
aguadoenzo, line 85: IIRC we don't mention the concept of `publishing` applications yet, maybe saying nothing is simpler?

getting-started.md
```diff
@@ -48,43 +46,51 @@ Once it is up and running you can get a shell inside the container with:
...
 
-    amc application publish <app name> 0
+    amc application show <app name>
 
-Now you can simply start a container for your new application by launching the application:
+If it says `published: true`, you can start a container for your new application by launching the application:
```
aguadoenzo, line 89: On new installs, `auto-publish` will always be true, we can assume that all created applications will be published

manage-images.md
```diff
@@ -0,0 +1,120 @@
+An image is the base for a container running in the Anbox Cloud. It contains all necessary components, like Anbox or the Android root file system. Each release of Anbox Cloud comes with an update image, which must be uploaded to the AMS service to make it available to be used inside any container.
```
aguadoenzo, line 1:
```suggestion
An image is the base for a container running in the Anbox Cloud. It contains all necessary components, like Anbox or the Android root file system. Each release of Anbox Cloud comes with an updated image, which must be uploaded to the AMS service to make it available to be used inside any container.
```

aguadoenzo, line 1: Images are automatically downloaded by AMS if the image server is configured (which we can assume it always is). The `which must be uploaded to the AMS service` sounds like the user has to do some manual stuff IMO

manage-images.md
aguadoenzo: @ru-fu that's correct. The delivery mechanism was more manual before, but the introduction of the image server changed things quite a bit

manage-images.md
```diff
@@ -0,0 +1,120 @@
...
+
+AMS can manage different types of images, which can be individually selected by applications. For that, each image must be named when it's initially uploaded to AMS. If the same image receives an update with the next release of Anbox Cloud, it does not need to be re-created but can simply be updated. This will cause all applications using the image to be updated and rebased on the new image version.
```
aguadoenzo, line 3: As mentioned above, images already have a name when they are downloaded by AMS, the user doesn't have to do anything.

Unrelated, but would it be worth mentioning that the *types* of images depend on the android version as well as the system architecture or would that be too much details ?

manage-images.md
aguadoenzo: Worth checking with @morphis but I feel like we should keep these as long as we mark them unsupported

manage-images.md
```diff
@@ -0,0 +1,120 @@
...
++----------------------+-----------------------------+--------+----------+
+```
+
+In contrast to manually added images, synchronised ones are marked as immutable as they can't be manually updated. In case you want to remove an immutable image, call the following command:
+
+    amc image delete --force io.anbox-cloud:nougat:amd64
+
+If you're not using `--force`, the command will fail.
+
+## Add a new image to AMS
```
aguadoenzo, line 52: Considering manual upload is rarely bound to happen, I think we should either drop this section or move it down, as it's less important than the rest

manage-images.md
aguadoenzo: I think the previous sentence misspelled `has` as `as` so your update is correct :)

web-dashboard.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Usage
+
+To access the dashboard, go to `https://<your-machine-address>/`.
+
+> **Hint**: The dashboard uses self-signed certificates. You might see a warning from your browser and have to accept the certificates manually.
+
+## Granting access
+
+Instead of implementing its own user management, the dashboard relies on OAuth for user authentication. The only OAuth provider supported right now is [Ubuntu One](https://login.ubuntu.com/).  Future versions of Anbox Cloud will allow using other OAuth providers.
```
aguadoenzo, line 22:
```suggestion
Instead of implementing its own user management, the dashboard relies on OAuth for user authentication. The only OAuth provider supported right now is [Ubuntu One](https://login.ubuntu.com/). Future versions of Anbox Cloud will allow using other OAuth providers.
```

aguadoenzo (changes requested):

manage-images.md
ru-fu: Seems there's quite a bit of outdated information in here. I understood it as if users had to do something - will update.

manage-images.md
ru-fu: We should have some more information on instance types, what they mean and how you choose the suitable one - but maybe that should go into the Instance types topic.
Or is there an "image type" in addition to an "instance type"?

manage-images.md
ru-fu: I made a new section for manual upload now and moved the "Add" and "Update" sections in there. I assume the "Default image" section applies in general, thus also for the image server? How about the "Delete" section - can you delete images/versions of images provided by the image server, or only those that you upload manually?

manage-images.md
ru-fu: Oh, that suddenly makes sense. :grin:

getting-started.md
aguadoenzo: Both `amc application ls` and `amc application show <id>` print the status of the application. However, `amc application ls` prints this information in a, arguably easier to read, table format, whereas `amc application show` prints yaml (with more information)

As for the `watch` command, it's just a way to avoid the user having to manually run the command every few seconds to check the status of the application, we don't necessarly have to mention it, it's meant to be a quality of life thing

manage-images.md
aguadoenzo: Good point, we have a reference for instance types but it's a bit lacking

In depends on what you mean by "image type". The different images we have are based on different android versions and different architectures (amd64, arm64) so you could say those are different types of images. However they are completely separate from instance-types

manage-images.md
aguadoenzo: Nice !

Default image rules apply to all images, regardless of where they come from
Deleting images is very similar with the exception that images coming from the image server are marked as `immutable`. This means that deleting them requires the `--force` flag. Aside from that, they're the same

aguadoenzo (approved):
LGTM :shipit:

manage-images.md
ru-fu: OK, now I understand your original comment. Let's get rid of the term "types" here, it's confusing.

manage-images.md
ru-fu: Oh right - that is actually documented, but in the wrong location. Moved it to the delete section now.

PR #15, opened 2021-09-07
add images

ru-fu:
Might not be a bad idea to keep a copy of our images in the repo
as well, just in case Discourse does something unexpected.
For now, adding the images from the AWS installation page that
I just archived.


ru-fu:
Will include this commit in the PR to update the installation docs.

PR #17, opened 2021-09-09
Appliance aws installation

ru-fu:


install-appliance.md
```diff
@@ -22,45 +22,70 @@ The Anbox Cloud Appliance is available through the AWS Marketplace and as a snap
...
 
-Installing the Anbox Cloud Appliance through the AWS Marketplace simplifies the installation and deployment process and allows billing to be handled directly through AWS. The following steps guide you through all relevant steps to deploy the Anbox Cloud Appliance in your AWS account.
+Installing the Anbox Cloud Appliance through the AWS Marketplace simplifies the installation and deployment process and allows billing to be handled directly through AWS.
 
-#### Choose an instance architecture
-
-First, decide if you want to deploy the Anbox Cloud Appliance on an x86 or Arm instance architecture. AWS supports both.
-
-The decision should factor in the following aspects:
+AWS supports running the Anbox Cloud Appliance on an x86 or Arm instance architecture. Before installing the appliance, decide which architecture you want to use. The decision should factor in the following aspects:
```
morphis, line 27: For the sake of marketing lets say:

```suggestion
AWS supports running the Anbox Cloud Appliance on x86 or the [AWS Graviton](https://aws.amazon.com/ec2/graviton/) Arm based instances. Before installing the appliance, decide which architecture you want to use. The decision should factor in the following aspects:
```

install-appliance.md
```diff
@@ -22,45 +22,70 @@ The Anbox Cloud Appliance is available through the AWS Marketplace and as a snap
...
 
 * GPUs are currently available for x86. NVIDIA GPUs will only become available for Arm instances [later in 2021](https://aws.amazon.com/blogs/machine-learning/aws-and-nvidia-to-bring-arm-based-instances-with-gpus-to-the-cloud/).
 * Not all Android applications support the x86 ABI. Therefore, some applications can run only on Arm.
 
-Once you have decided for an instance architecture, go to the corresponding marketplace page:
+For detailed information about the offering, see the following pages on the AWS Marketplace:
 
-* [Anbox Cloud Appliance for x86](https://aws.amazon.com/marketplace/pp/prodview-3lx6xyaapstz4)
 * [Anbox Cloud Appliance for Arm](https://aws.amazon.com/marketplace/pp/prodview-aqmdt52vqs5qk)
```
morphis, line 34: See above

install-appliance.md
```diff
@@ -116,9 +156,13 @@ The appliance requires a few additional tools.
 
 After the installation, access `https://your-machine-address`. This web page provides status information for the following initialisation process.
 
+> **NOTE:** By default, the Anbox Cloud Appliance uses self-signed certificates, which might cause a security warning in your browser. Use the mechanism provided by your browser to proceed to the web page.
```
morphis, line 159: @aguadoenzo should we say something on how to do this on Chrome on Mac?

install-appliance.md
```diff
@@ -70,15 +95,30 @@ In this example, we use three storage volumes:
...
 
-You should now review the instance configuration. If everything is correct, launch the instance.
+You should now review the instance configuration. If everything is correct, click **Launch**.
+
+You are prompted to select a key pair. You can choose an existing key pair or create one if you don't have one yet. Make sure to save the private key in a secure location.
+
+<!-- add image -->
+![Confirm to launch instances](images/install_appliance_launch-instances.png)
+
+Click **Launch Instances** to continue. AWS will verify your configuration, subscribe you to the product and launch the instance.
```
morphis, line 118: Hasn't subscribing happened way before this?

install-appliance.md
```diff
@@ -22,45 +22,70 @@ The Anbox Cloud Appliance is available through the AWS Marketplace and as a snap
...
-Once you have decided for an instance architecture, go to the corresponding marketplace page:
+For detailed information about the offering, see the following pages on the AWS Marketplace:
 
-* [Anbox Cloud Appliance for x86](https://aws.amazon.com/marketplace/pp/prodview-3lx6xyaapstz4)
 * [Anbox Cloud Appliance for Arm](https://aws.amazon.com/marketplace/pp/prodview-aqmdt52vqs5qk)
+* [Anbox Cloud Appliance for x86](https://aws.amazon.com/marketplace/pp/prodview-3lx6xyaapstz4)
+
+The following instructions guide you through all relevant steps to deploy the Anbox Cloud Appliance in your AWS account. For additional information, see the [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/launching-instance.html) about launching an instance.
+
+#### 1. Start the launch wizard
```
morphis, line 39: Don't you need to subscribe to the marketplace offering before?

morphis (changes requested):

install-appliance.md
ru-fu: Is that not confusing, since the page itself just says Arm?

install-appliance.md
ru-fu: I suggest to add a more detailed explanation to https://anbox-cloud.io/docs/manage/web-dashboard (since this doesn't apply only to the appliance) and link to that section from here (once we have the section).

install-appliance.md
ru-fu: No. If you start with subscribing, you will click on "Continue to configure" next and end up in the wizard that doesn't let you add storage.

install-appliance.md
ru-fu: No. If you start without a subscription, you see the prices when you select the AMI (but you just click "continue") and then the real subscribing happens when you launch the instance.

install-appliance.md
```diff
@@ -116,9 +156,13 @@ The appliance requires a few additional tools.
 
 After the installation, access `https://your-machine-address`. This web page provides status information for the following initialisation process.
 
+> **NOTE:** By default, the Anbox Cloud Appliance uses self-signed certificates, which might cause a security warning in your browser. Use the mechanism provided by your browser to proceed to the web page.
+
 ![appliance-welcome|690x343, 100%](upload://yIGZThPljsjPyRAVQVFkZOiVVNF.png)
```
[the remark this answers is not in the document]
ru-fu: Yes, those images are handled by Discourse, and we don't have them in the repo yet. I only added the new pictures. So it's expected that the unchanged ones don't work here.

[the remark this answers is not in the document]
adglkh: Okay. I see.

install-appliance.md
aguadoenzo: Agreed with @ru-fu. Alternatively we could link to a 3rd party blog post explaining this but I'm not sure this is the best idea

install-appliance.md
```diff
@@ -116,9 +156,13 @@ The appliance requires a few additional tools.
...
 
-Invoke the initialisation process of the Anbox Cloud Appliance via SSH on the machine that hosts the appliance:
+Log on to the machine that hosts the appliance. If you installed on an AWS instance, see [Connect to your Linux instance using SSH](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AccessingInstancesLinux.html) for instructions.
```
aguadoenzo, line 163: This link might confuse users as the Ubuntu image doesn't have an `ec2-user` user, just the `ubuntu` user. We should either tell the user to simply run `ssh ubuntu@<machine-ip>` or point to that article but specify that the user should be `ubuntu` instead of `ec2-user`

aguadoenzo (changes requested):

install-appliance.md
ru-fu: Just "ssh ubuntu@<machine-ip>" doesn't work, because you need to specify the key file. Not sure if that can be prevented by naming the file differently? 
I definitely needed the help article to figure out what to do. But sure, I can point out the user name.

aguadoenzo: On ubuntu renaming the key to `id_rsa` does the trick, or you can use `ssh -i $HOME/.ssh/aws-ssh-key.pem ubuntu@foobar.com`
Let's leave the article then, with a hint to the proper username

install-appliance.md
ru-fu: We have a task to track this, so let's keep that update separate from this PR.

install-appliance.md
ru-fu: Added information about the user name and key path. Seems the only way to get it to work without specifying the path is to rename the key file, and that is not good usability since everybody will already have a file with that name in the .ssh folder ... :(

aguadoenzo (approved):

install-appliance.md
morphis: Yeah, I am working torwards fixing that too.

install-appliance.md
morphis: Ok

morphis (approved):

PR #19, opened 2021-09-10
fix clustering documentation

ru-fu:
Combine capacity planning and LXD auto scaling into one "About
clustering" page and split out how-tos for scaling up and down.

The first two paragraphs in clustering.md are new. The rest
is existing content with some language fixes.


clustering.md
```diff
@@ -0,0 +1,82 @@
+While it is possible to install Anbox Cloud on a single machine, you usually want to distribute the load over several machines in a cluster. In such a clustering setup, one node is dedicated to host the AMS service (the management node). If you use the streaming stack, two additional nodes are dedicated to host the extra services required for streaming. All other nodes are used as worker nodes.
+
+Each worker node runs [LXD](https://linuxcontainers.org/) in clustering mode, and this LXD cluster is used to host the Android containers.
+
+## Capacity planning
+
+Anbox Cloud is optimised to provide containers at high density per host. However, to provide enough underlying resources for a specific number of containers, we need to do some calculations to find out how many LXD machines (thus worker nodes) with how many resources we need.
```
morphis, line 7:
```suggestion
Anbox Cloud is optimised to provide containers at high density per host. However, to provide enough underlying resources for a specific number of containers, you need to do some calculations to find out how many LXD machines (thus worker nodes) with how many resources you need.
```

clustering.md
```diff
@@ -0,0 +1,82 @@
...
+```bash
+CPU allocation rate = 100 * 2 CPU cores / 48 CPU cores ~= 5
+Memory needed = 100 * 3 GB / 2 = 150 GB
+Disk space needed = 100 * 3 GB = 300 GB
+```
+
+In this example, we used a memory allocation rate of `2`.
+
+Which CPU allocation rate makes sense always depends on which type of application will be running inside the containers and which amount of CPU it needs. For low CPU intensive applications a higher and for high CPU intensive applications a lower allocation rate makes sense.
+
```
morphis, line 52: I think we need to maybe integrate what we have about `gpu-slots` and `gpu-encoder-slots` here too (see https://anbox-cloud.io/docs/manage/gpu-support) or at least link to it

ru-fu: Thanks, added!

morphis (approved):
LGTM

PR #20, opened 2021-09-10
fix application documentation

ru-fu:
Split the content into different how-to, reference, and
explanation pages. No new content, just moved the content around
with some minor cleanup.


create-application.md
```diff
@@ -0,0 +1,121 @@
...
+resources:
+  memory: 4GB
+  disk-size: 8GB
+```
+
+Once the status of the application switches to `ready`, the application is ready and can be used. See [Wait for an application](tbd) for information about how to monitor the application status.
+
+## Create from a tarball
+
+An application can also be created from a tarball file. The tarball file must be compressed with bzip2 and must follow the format of the application package as described before.
```
ru-fu, line 109: What do we mean with "as described before"? The folder structure from line 50-52?

morphis: Yes

morphis (approved):

PR #29, opened 2021-09-28
feat(addons): update addons documentation

aguadoenzo:
With v2 hooks (pre-start, post-start, post-stop) introduced, the old hooks are deprecated. This commit documents the new hooks and separates topics in a tutorial, a reference and a how-to guide


addons-reference.md
```diff
@@ -0,0 +1,58 @@
+This document provides a complete reference on addons in Anbox Cloud.
+
+<h2 id='heading--addon-reference-file-structure'>File structure</h2>
```
ru-fu, line 3: I've started using
```
<a name="short-name"></a>
## Heading
```
to add an anchor. The advantage is that it's easier to change the heading level and to remember the anchor name, but the disadvantage is that it adds a bit of empty space.
Both should work though.

addons-reference.md
```diff
@@ -0,0 +1,58 @@
...
+When creating or updating an addon, the directory containing your addon files **MUST**:
+
+- Contain a file named `manifest.yaml`.
+- Contain a directory named `hooks`.
+    - `hooks` **MUST** contain at least one executable file with a valid hook name (see *Hooks* below)
+
+Other files are bundled with the addon and can be accessed in hooks under `$ADDON_DIR` (see [*Environment variables*](#heading--addon-reference-env-variables)).
+```bash
+cat "$ADDON_DIR"/public_key.pem >> ~/.ssh/authorized_keys
+```
```
ru-fu, line 13: What is this about?

addons-tutorial.md
```diff
@@ -0,0 +1,91 @@
+Addons are used to customize the images used for the containers. An addon has *hooks* that are
+invoked at various points in the lifecycle of a container. A good example is an addon which
+enables the SSH service inside a container to provide access from the outside for any kind
+of automation of the container.  
+
+Addons are created independently, and can be attached to individual applications.
```
ru-fu, line 6: I think this will work nicely as a tutorial, great idea! :)
We don't have any "real" tutorials in the docs yet, so there isn't much to refer to. But overall, we don't want a lot of explanation in here - we just want to guide users through the steps to get a result.

So the introduction should be a bit more direct. Not explaining why addons are nice and what you can do with them, but telling users that in this tutorial, they will create an addon with a hook that enables the SSH service.

addons-tutorial.md
```diff
@@ -0,0 +1,91 @@
...
+
+<h2 id='heading--build-your-own-addon'>Build your first addon</h2>
```
ru-fu, line 8: This heading isn't needed imo. That is what the whole tutorial is about. :wink: 

I suggest to use numbered headings here, one for every step that the user should do. So:
1. Create a manifest file
2. Add a pre-start hook
3. and 4. as is

addons-tutorial.md
```diff
@@ -0,0 +1,91 @@
...
+
+Any addon must have a `manifest.yaml` file and at least one hook in the `hooks` directory, e.g:
+
+```text
+manifest.yaml
+hooks/pre-start
+```
```
ru-fu, line 15: Not needed here - we have it in the reference already, and the user will see it more clearly after creating the files.

addons-tutorial.md
```diff
@@ -0,0 +1,91 @@
...
+#!/bin/bash
+
+systemctl start sshd
+```
+
+> INFO: supported hooks are: `pre-start`, `post-start` and `post-stop`. Read more about them [here](TODO: INSERT LINK TO ADDON REFERENCE)
+
+### Create the addon
+
+Assuming your addon structure currently looks like this:
```
ru-fu, line 40: No need to assume - this is what we have created. :wink: 
It makes sense to show the structure here imo, but introduce with something like "This is the addon structure you  created". You could also add a link to the file structure section in the reference here.

addons-tutorial.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+The `amc wait` command returns when your application is ready to launch.  
+You can now launch an instance of your application.
+```bash
+amc launch -s +ssh my-application
+```
+> NOTE: port 22 (ssh) is closed by default. We open it via services. Read more about services [here](https://anbox-cloud.io/docs/howto/container/expose-services)
+
+You should now be able to access your container via ssh.
```
ru-fu, line 83:
```suggestion
You can now access your container via ssh.
```
We shouldn't expect anything to go wrong in a tutorial. ;)

howto-addons.md
```diff
@@ -0,0 +1,118 @@
+This section explores more advanced use cases with addons. For an introduction, see [the addon tutorial](TODO: link to the tutorial page)
```
ru-fu, line 1: You suggested to have a separate section for addons, and I think that makes a lot of sense. We didn't have it yet since the current addon page had all how-tos on one page, and I hadn't cleaned it up yet. 

I suggest to have a wrapper page for addons that should contain a short explanation on what they are and what they should be used for, plus links to the tutorial and reference. And then each of the sections below can be a separate how-to - except for the Best practices, those don't really work as how-to. For now, I think we could just put them on the wrapper page. We can split them out later if we end up with an explanation page for addons.

ru-fu (commented):
This looks really good! :-)
I added some general comments about the structure. I haven't reviewed the language yet (a lot of it is probably copied over from the old docs anyway), but I can do that when the overall structure is in place.
Let me know if you want me to add a commit with some or all of the changes.

addons-reference.md
aguadoenzo: This is to showcase an example of an addon accessing a file that was bundled at its creation, I'll add a few steps to make it more comprehensible

ru-fu:
@aguadoenzo I merged the updated file structure now. It shouldn't give you any conflicts, but please rename your files accordingly:

addons-reference.md => reference/addons.md
addons-tutorial.md => tutorial/creating-addon.md
howto-addons.md => howto/addon/landing.md, howto/addon/update.md, ...
manage-addons.md => delete

howto/addon/apply-to-all-applications.md
```diff
@@ -0,0 +1,13 @@
...
+
+However, if you want to use the same addon or addons for all your applications, you can enable them globally. To do so, run the following command after creating your addons:
+
+```bash
+amc config set application.addons foo,bar
+```
+
+This command adds the `foo` and `bar` addons to all your new and existing applications. AMS will automatically update existing applications.
+
+Global addons will be added to application-specific addons.
```
ru-fu, line 11: What does this sentence mean?

howto/addon/customize-android.md
```diff
@@ -0,0 +1,14 @@
+If you need to do any customisation to the Android system that runs your application, you can use the `anbox-shell` tool within a hook. This tool is useful to interact with the Android system directly.
+
+The following hook configures the Android system to use an HTTP proxy:
```
ru-fu, line 3: What kind of hook is this? post-start?

howto/addon/apply-to-all-applications.md
ru-fu: Now it suddenly makes perfect sense, but I'll reword it to make it clearer. ;)

tutorial/creating-addon.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+[note type="information" status="Note"]The SSH port 22 is closed by default. Therefore, we open it by [exposing its service](https://anbox-cloud.io/docs/howto/container/expose-services).[/note]
+
+You can now access your container via SSH:
+```bash
+ssh -i ~/ssh-addon-key root@<container_ip> -p <exposed port>
+```
+
+[note type="information" status="Note"] The exposed port can be found be running `amc ls`, under the `ENDPOINTS` column. Exposed ports usually start around port 10000.
```
ru-fu, line 86:
```suggestion
[note type="information" status="Note"] The exposed port can be found be running `amc ls`, under the `ENDPOINTS` column. Exposed ports usually start around port 10000.[/note]
```

PR #32, opened 2021-09-29
update URL structure and add redirects for old URLs

ru-fu:


morphis (approved):
LGTM

PR #33, opened 2021-09-30
rework index page and overview

ru-fu:
- Moving some content from the index page to the overview
- Adding links to important information to the start page
- Expanding the information about Anbox Cloud vs. Anbox Cloud Appliance

Signed-off-by: Ruth Fuchss <ruth.fuchss@canonical.com>

PR #36, opened 2021-10-04
Drop all references to oem-share

morphis:
Nobody has access to oem-share outside of Canonical at this point so
linking to it doesn't make sense. This uses a static link to the 1.10
platform SDK API but we should look for a better way to manage the
different references we have to it.

### DIFF
--- a/explanation/anbox-cloud.md
+++ b/explanation/anbox-cloud.md
@@ -32,7 +32,7 @@ A developer or system administrator will manage **AMS** through the **command li
 
 For example, a simple Android application testing service would provide a user-facing interface dealing with things like authentication and user management, and would communicate with the REST API to add applications or start and stop containers when a user asks to.
 
-Anbox Cloud can be heavily customized and extended via [**Platform Plugins**](https://oem-share.canonical.com/partners/indore/share/docs/1.7/en/sdk/anbox/) and [**addons**](https://discourse.ubuntu.com/t/managing-addons/17759). Platform plugins and addons can be built to add specific streaming capabilities, perform operations within Android containers and much more. One example of a platform plugin is the **Anbox WebRTC Platform** used in the Anbox Streaming Stack. Addons are ways to customize the base image by installing additional software and running scripts on different lifecycle hooks.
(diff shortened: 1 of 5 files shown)


explanation/anbox-cloud.md
```diff
@@ -32,7 +32,7 @@ A developer or system administrator will manage **AMS** through the **command li
...
 
-Anbox Cloud can be heavily customized and extended via [**Platform Plugins**](https://oem-share.canonical.com/partners/indore/share/docs/1.7/en/sdk/anbox/) and [**addons**](https://discourse.ubuntu.com/t/managing-addons/17759). Platform plugins and addons can be built to add specific streaming capabilities, perform operations within Android containers and much more. One example of a platform plugin is the **Anbox WebRTC Platform** used in the Anbox Streaming Stack. Addons are ways to customize the base image by installing additional software and running scripts on different lifecycle hooks.
+Anbox Cloud can be heavily customized and extended via [**Platform Plugins**](https://anbox-cloud.github.io/1.10/anbox-platform-sdk/index.html) and [**addons**](https://discourse.ubuntu.com/t/managing-addons/17759). Platform plugins and addons can be built to add specific streaming capabilities, perform operations within Android containers and much more. One example of a platform plugin is the **Anbox WebRTC Platform** used in the Anbox Streaming Stack. Addons are ways to customize the base image by installing additional software and running scripts on different lifecycle hooks.
```
ru-fu, line 35:
```suggestion
Anbox Cloud can be heavily customized and extended via [**Platform Plugins**](https://discourse.ubuntu.com/t/anbox-cloud-sdks/17844#anbox-platform-sdk) and [**addons**](https://discourse.ubuntu.com/t/managing-addons/17759). Platform plugins and addons can be built to add specific streaming capabilities, perform operations within Android containers and much more. One example of a platform plugin is the **Anbox WebRTC Platform** used in the Anbox Streaming Stack. Addons are ways to customize the base image by installing additional software and running scripts on different lifecycle hooks.
```
That's what I changed it to in https://github.com/anbox-cloud/discourse-docs/pull/33, and I think this gives a better intro to platform plugins than just the API.

ru-fu (changes requested):

aguadoenzo (approved):

PR #54, opened 2021-11-05
clean up Deploy with Juju page

ru-fu:
Some general cleanup. This isn't detailed enough to be a
tutorial, so make sure it works as howto.

### DIFF
--- a/howto/install/deploy-juju.md
+++ b/howto/install/deploy-juju.md
@@ -14,6 +14,7 @@ Before you start the installation, ensure that you have the required credentials
 * Account credentials for one of the following public clouds:
   * [Amazon Web Services](https://aws.amazon.com/) (including AWS-China)
   * [Google Cloud platform ](https://cloud.google.com/)
+  * [Microsoft Azure](https://azure.microsoft.com/)
 * Your *Ubuntu Advantage for **Applications*** token. If you don't have one yet, [speak to your Canonical representative](https://anbox-cloud.io/contact-us). If you already have a UA Applications token, sign in on https://ubuntu.com/advantage to retrieve it.
   [note type="caution" status="Warning"]The *Ubuntu Advantage for **Infrastructure*** token that every user gets for free for personal use does **NOT** work and will result in a failed deployment. You must purchase a *Ubuntu Advantage for **Applications*** subscription by [contacting Canonical](https://anbox-cloud.io/contact-us).[/note]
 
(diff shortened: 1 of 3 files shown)


howto/install/deploy-juju.md
```diff
@@ -1,73 +1,67 @@
-With this quick start guide and some tools from Canonical, you'll have an Anbox Cloud running on the cloud of your choice in minutes!
+Anbox Cloud supports various public clouds, such as AWS, Azure and Google. To deploy Anbox Cloud in a cloud environment, you use Juju.
```
ru-fu, line 1: Is it correct that Azure is supported? It was mentioned below, but it isn't in the prerequisites.

howto/install/deploy-juju.md
```diff
@@ -151,15 +151,15 @@ machines:
     constraints: "instance-type=m4.2xlarge root-disk=50G"
 ```
 
-To deploy with the bundle from above, add `--overlay overlay.yaml` to your deploy command:
+To deploy, add `--overlay overlay.yaml` to your deploy command. For example:
 
     juju deploy cs:~anbox-charmers/anbox-cloud ... --overlay overlay.yaml
```
ru-fu, line 156: Does the "..." mean that you might need to add something? Probably the `--overlay ua.yaml`?
In this case, wouldn't it make more sense to just add this here, since most (all?) users will need it and this is only an example anyway?

howto/install/deploy-juju.md
```diff
@@ -174,13 +174,13 @@ machines:
...
 
-### Use Arm Instances
+### Use Arm instances
 
-Some clouds, like AWS with their Graviton instances, provide support for Arm instance types. These can be used with Anbox Cloud by specifying the correct instance type in the `overlay.yaml` as well:
+Some clouds, like AWS with their Graviton instances, provide support for Arm instance types. These can be used with Anbox Cloud by specifying the correct instance type in the `overlay.yaml`:
 
 ```
 applications:
```
ru-fu, line 186: Question for line 188 below: I thought we don't support Juju 2.9, so why that comment - and does this syntax work with Juju 2.8?

morphis (approved):
LGTM

howto/install/deploy-juju.md
morphis: As of 1.12 we support 2.9 just fine. However not sure which comment you mean

morphis (approved):

PR #63, opened 2021-12-01
feat(anbox-http-api): support to carry gps accuracy info on gps data update

adglkh:


reference/anbox-http-api.md
```diff
@@ -146,42 +146,15 @@ Input:
...
-    ]
-  },
-  "status": "Success",
-  "status_code": 200,
-  "type": "sync"
-}
-
+    "format": "wgs84",               # (optional) Location format; possible values are "nmea" or "wgs84". Defaults to "wgs84" if not specified
+    "horizontal_accuracy": 20,       # (optional) horizontal accuracy in meters. Defaults to 20 if not specified
+    "vertical_accuracy": 20          # (optional) vertical accuracy in meters. Defaults to 20 if not specified
```
ru-fu, line 157: There's a note just above the code block that explains the format - can you add the explanation there instead of in a comment?

adglkh: Yes, I can. However, I feel it looks lengthy if we move the meaning of each field from the comment to the area which is wrapped by note syntax. Having them in comments looks more structured and easy to figure out what's the meaning of each field for readers. (Maybe I was wrong :))
How do you feel about this? 
Meanwhile, I just found in some places we are using **// <-** in the code block to explain the meaning of one field in the JSON output.  I need to fix that for consistency.

ru-fu: I think it's okay to use comments to explain what a field contains (like "Latitude of geographic coordinates") - but when it comes to explaining different options and default values, it's hard to be clear enough in just a comment. And especially in this case where we have different options, and the code only shows one of them (namely wgs84).
So I suggest to keep the comments short and precise, and add additional information in the main docs.

reference/anbox-http-api.md
```diff
@@ -140,48 +140,21 @@ $ curl -s -X POST --unix-socket /run/user/1000/anbox/sockets/api.unix s/1.0/loca
...
+[note type="information" status="Note"]The latitude or longitude of geographic coordinates is expressed in [decimal degree](https://en.wikipedia.org/wiki/Decimal_degrees) form (WGS84 data format) as shown below in the example, whereas the NMEA-based data format is expressed in [ddmm.mm](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion) (d refers to degrees, m refers to minutes). No matter which format you use, northern latitudes or eastern longitudes are positive, southern latitudes or western longitudes are negative. The location format can be either "nmea" or "wgs84". If it's omitted, this attribute’s value defaults to "wgs84". Both vertical and horizontal accuracy are measured in meters, the default value for GPS accuracy is 20 meters. [/note]
```
ru-fu, line 143:
```suggestion
[note type="information" status="Note"]The latitude or longitude of geographic coordinates can be expressed in [decimal degree](https://en.wikipedia.org/wiki/Decimal_degrees) form (WGS84 data format) as shown below in the example or in an NMEA-based data format as [ddmm.mm](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion) (d refers to degrees, m refers to minutes). Specify the format by setting the `format` field to either `"wgs84"` or `"nmea"`. If the field is omitted, its value defaults to `"wgs84"`. No matter which format you use, northern latitudes or eastern longitudes are positive, southern latitudes or western longitudes are negative. 

Both vertical and horizontal accuracy are measured in meters. The default value for GPS accuracy is 20 meters. [/note]
```

reference/anbox-http-api.md
```diff
@@ -140,48 +140,21 @@ $ curl -s -X POST --unix-socket /run/user/1000/anbox/sockets/api.unix s/1.0/loca
...
-    ]
-  },
-  "status": "Success",
-  "status_code": 200,
-  "type": "sync"
-}
-
+    "format": "wgs84",               # (optional) Location format
+    "horizontal_accuracy": 20,       # (optional) horizontal accuracy in meters
+    "vertical_accuracy": 20          # (optional) vertical accuracy in meters
```
ru-fu, line 157:
```suggestion
    "horizontal_accuracy": 20,       # (optional) Horizontal accuracy in meters
    "vertical_accuracy": 20          # (optional) Vertical accuracy in meters
```

ru-fu (approved):

PR #65, opened 2021-12-02
add a note about G5g instances not being available in all regions

ru-fu:


morphis (approved):
LGTM, thanks!

PR #67, opened 2021-12-06
feat(reference): add benchmark numbers for new g5g.metal instance

morphis:


reference/perf-benchmarks.md
```diff
@@ -15,3 +15,7 @@ The following benchmarks give an overview of the performance that you can achiev
 |---------------|---------------------|----------------------|------------|---------------------|--------------|---------|
 | m6g.2xlarge   | 1.11.2 (appliance)  | software / software  | 720p       | Android 10 Launcher | 3            | 20      |
 | m5a.2xlarge   | 1.11.2 (appliance)  | software / software  | 720p       | Android 10 Launcher | 3            | 13      |
+| g5g.metal     | 1.12.1 (appliance)  | hardware / hardware  | 720p       | [BombSquad](https://www.froemling.net/apps/bombsquad) | 55 | 28 |
+| g5g.metal     | 1.12.1 (appliance)  | hardware / hardware  | 720p       | [BombSquad](https://www.froemling.net/apps/bombsquad) | 40 | 55 |
```
adglkh, line 19: What's the key factor which yields out different numbers of containers and avg FPS when rendering and encoding both are based on hardware and instance-type/appliance version are identical in the above two items.

reference/perf-benchmarks.md
```diff
@@ -15,3 +15,7 @@ The following benchmarks give an overview of the performance that you can achiev
...
+| g5g.metal     | 1.12.1 (appliance)  | hardware / hardware  | 1080p      | [BombSquad](https://www.froemling.net/apps/bombsquad) | 40 | 27 |
+| g5g.metal     | 1.12.1 (appliance)  | hardware / hardware  | 1080p      | [BombSquad](https://www.froemling.net/apps/bombsquad) | 20 | 57 |
```
adglkh, line 21: Same here.

adglkh (commented):
reviewed.

reference/perf-benchmarks.md
morphis: We have a different target FPS. In the first case the target is 30, in the second case it is 60. So the two lines show that you can run higher number of containers by accepting lower frame rates and running with higher frame rates when accepting a lower number of containers.

adglkh: Then we need a column like `Requested FPS` as another factor that people can differentiate those two items, otherwise it introduces confusion.

ru-fu: So "Avg. FPS" is a value that you specify?
This is a bit misleading. The way the table is ordered, and the fact that it's an average, makes it seem like a result. So that with the settings from the first 5 columns, you achieve the results in the last two columns.

morphis: The target FPS is more an implementation detail to stabilize the frame rate across a set of containers. The difference here is that the more clients per GPU you have the less FPS each of them can achieve. That's independent if we limit the FPS or not. So with 40 clients you can achieve higher frame rates than with 55. 

The "Avg FPS" column is the result, not an input. If we add another column specifying an input "Target FPS" I don't know if that actually helps to understand the results as in the end it's just an implementation detail to get more stablized FPS values across a set of clients an not one at 50 and one at 20.

ru-fu: I think we need to add something - because right now, the only thing you can see from the table is that you can get different results with the same setup, but you don't know how.

reference/perf-benchmarks.md
```diff
@@ -1,17 +1,34 @@
-The following benchmarks give an overview of the performance that you can achieve with Anbox Cloud. The numbers were collected on various virtual and bare metal hardware.
+The following benchmarks give an overview of the performance that you can achieve with Anbox Cloud.
+
+The benchmarks are performed using the `amc benchmark` utility as described [here](https://discourse.ubuntu.com/t/benchmarking-a-deployment/17770). The results describe the maximum number of parallel running containers (colum "# Containers") delivering a stable frame rate (column "Avg. FPS"). Running more containers either gives to high variation in the provided frame rate or is not possible due to other hardware limitions (system memory, GPU memory, ...).
```
ru-fu, line 3:
```suggestion
The benchmarks were performed using the `amc benchmark` utility as described in [About benchmarking](https://discourse.ubuntu.com/t/benchmarking-a-deployment/17770). The results describe the maximum number of parallel running containers (column "# Containers") delivering a stable frame rate (column "Avg. FPS"). Running more containers either gives too high variation in the provided frame rate or is not possible due to other hardware limitations (system memory, GPU memory, ...).
```

reference/perf-benchmarks.md
```diff
@@ -1,17 +1,34 @@
...
+
+All benchmark include rendering and video encoding. On machines/VMs without a GPU rendering and video encoding are performed in software on the CPU.
```
ru-fu, line 5:
```suggestion
All benchmarks include rendering and video encoding. On machines/VMs without a GPU, rendering and video encoding are performed in software on the CPU.
```

reference/perf-benchmarks.md
```diff
@@ -1,17 +1,34 @@
...
+
+For most of the benchmarks below a special version of the application [BombSquad](https://www.froemling.net/apps/bombsquad) is used. It runs in a demo mode where the game provides random and automated simulated gameplay. This allows the benchmark to simulate a real world scenario where actual users would play the game instead of sitting on a static game scene without much variation across the Android instances.
```
ru-fu, line 7:
```suggestion
For most of the benchmarks below, a special version of the [BombSquad](https://www.froemling.net/apps/bombsquad) application was used. This version runs in a demo mode in which the game provides random and automated simulated gameplay. This allows the benchmark to simulate a real-world scenario where actual users would play the game, instead of sitting on a static game scene without much variation across the Android instances.
```

reference/perf-benchmarks.md
```diff
@@ -1,17 +1,34 @@
...
+
+All benchmarks are done with a variation of the following `amc benchmark` command
```
ru-fu, line 9:
```suggestion
All benchmarks are done with a variation of the following `amc benchmark` command:
```

reference/perf-benchmarks.md
```diff
@@ -1,17 +1,34 @@
...
+
+    mc benchmark --network-address 192.168.100.1 -n <number of containers> \
```
ru-fu, line 11:
```suggestion
    amc benchmark --network-address 192.168.100.1 -n <number of containers> \
```

ru-fu (changes requested):
Thanks, that is a lot clearer! Just some language fixes.

ru-fu (approved):

PR #73, opened 2022-01-07
add a glossary

ru-fu:

### DIFF
--- a/reference/glossary.md
+++ b/reference/glossary.md
@@ -1,0 +1,451 @@
+[Details="Alphabetical list of terms"]
+- [AAM](#aam)
+- [AAR (Anbox Application Registry)](#aar)
+- [AAR (Android Archive)](#android-archive)
+- [ADB](#adb)
+- [AMC](#amc)
+- [AMS](#ams)
+- [AMS Node Controller](#ams-node-controller)
+- [AMS SDK](#ams-sdk)
+- [APK](#apk)
+- [AWS](#aws)
+- [Addon](#addon)
+- [Amazon Web Services](#aws)
+- [Anbox](#anbox)
+- [Anbox Application Manager](#aam)
+- [Anbox Application Registry](#aar)
+- [Anbox Cloud](#anbox-cloud)
+- [Anbox Cloud Appliance](#anbox-cloud-appliance)
+- [Anbox Management Client](#amc)
+- [Anbox Management Service](#ams)
+- [Anbox Platform SDK](#anbox-platform-sdk)
+- [Anbox Streaming SDK](#anbox-streaming-sdk)
+- [Android Archive](#android-archive)
+- [Android Debug Bridge](#adb)
+- [Android Package Kit](#apk)
+- [Android app](#android-app)
+- [Appium](#appium)
+- [Application](#application)
+- [Application container](#application-container)
+- [Application manifest](#application-manifest)
+- [Base container](#base-container)
+- [Boot package](#boot-package)
+- [Bootstrap process](#bootstrap-process)
+- [Container](#container)
+- [Core stack](#core-stack)
+- [Coturn](#coturn)
+- [GPU](#gpu)
+- [Grafana](#grafana)
+- [Graphics Processing Unit](#gpu)
+- [HA](#ha)
+- [High availability](#ha)
+- [Hook](#hook)
+- [Image](#image)
+- [Instance type](#instance-type)
+- [Juju](#juju)
+- [LXD](#lxd)
+- [Manifest](#application-manifest)
+- [Monitoring stack](#monitoring-stack)
+- [NATS](#nats)
+- [Nagios](#nagios)
+- [Neural Autonomic Transport System](#nats)
+- [Node controller](#ams-node-controller)
+- [Platform](#platform)
+- [Prometheus](#prometheus)
+- [Raw container](#raw-container)
+- [Regular container](#regular-container)
+- [STUN/TURN server](#stun/turn-server)
+- [Scrcpy](#scrcpy)
+- [Session](#session)
+- [Snap](#snap)
+- [Software Rasterization](#swrast)
+- [Stream agent](#stream-agent)
+- [Stream gateway](#stream-gateway)
+- [Streaming stack](#streaming-stack)
+- [Swrast](#swrast)
+- [Ubuntu Advantage for Applications](#ubuntu-advantage-for-applications)
+- [Ubuntu One](#ubuntu-one)
+- [Watchdog](#watchdog)
+- [Web dashboard](#web-dashboard)
+- [WebRTC](#webrtc)
+
+[/Details]
+
+## Definitions
+
+<a name="addon"></a>
+### Addon
+
+A piece of code that can be used to extend and customise images in Anbox Cloud.
+
+See https://discourse.ubuntu.com/t/addons/25293.
+
+<a name="aws"></a>
+### Amazon Web Services (AWS)
+
+A cloud platform provided by Amazon that can be used to host Anbox Cloud.
+
+See [the AWS website](https://aws.amazon.com/).
+
+<a name="ams-node-controller"></a>
+### AMS Node Controller
+
+A service that runs on every LXD node and puts the appropriate firewall rules in place when a container is started or stopped.
+
+<a name="ams-sdk"></a>
+### AMS SDK
+
+An SDK that provides Go language bindings for connecting to AMS through the exposed REST API.
+
+See [AMS SDK](https://discourse.ubuntu.com/t/anbox-cloud-sdks/17844#ams-sdk).
+
+<a name="anbox"></a>
+### Anbox
+
+A component of Anbox Cloud that facilitates booting an Android system on a regular GNU/Linux system. The concepts of the Anbox component in Anbox Cloud are similar to the [Anbox open source project](https://github.com/anbox/anbox), but the Anbox open source project is an independent project that is not related to or used in Anbox Cloud.
+
+<a name="aam"></a>
+### Anbox Application Manager (AAM)
+
+A utility (`aam`) that is installed in the Anbox image and that can be used for various tasks, for example, to back up and restore Android application data.
+
+See https://discourse.ubuntu.com/t/back-up-and-restore-application-data/24183.
+
+<a name="aar"></a>
+### Anbox Application Registry (AAR)
+
+A central repository for applications created in Anbox Cloud. Using an AAR is very useful for larger deployments to keep applications in sync.
+
+See https://discourse.ubuntu.com/t/anbox-application-registry-aar/17761.
+
+<a name="anbox-cloud"></a>
+### Anbox Cloud
+
+A rich software stack that enables you to run Android applications in the cloud for all kinds of different use cases, including high-performance streaming of graphics to desktop and mobile client devices.
+
+See https://discourse.ubuntu.com/t/about-anbox-cloud/17802.
+
+<a name="anbox-cloud-appliance"></a>
+### Anbox Cloud Appliance
+
+A self-contained, single-machine deployment variant of Anbox Cloud.
+
+See [Variants](https://discourse.ubuntu.com/t/about-anbox-cloud/17802#variants).
+
+<a name="amc"></a>
+### Anbox Management Client (AMC)
+
+The command line interface that is used to manage the Anbox Management Service (AMS).
+
+<a name="ams"></a>
+### Anbox Management Service (AMS)
+
+The service that handles all aspects of the application and container life cycle in Anbox Cloud. AMS is responsible for managing containers, applications, addons, updates and more, ensuring high density, performance and fast container startup times.
+
+AMS uses [etcd](https://etcd.io/) as database. It connects to LXD over its REST API.
+
+See https://discourse.ubuntu.com/t/anbox-management-service-ams/24321.
+
+<a name="anbox-platform-sdk"></a>
+### Anbox Platform SDK
+
+A C/C++ SDK that provides support for developing custom platform plugins, which allows users to integrate Anbox with their existing infrastructure.
+
+See [Anbox Platform SDK](https://discourse.ubuntu.com/t/anbox-cloud-sdks/17844#anbox-platform-sdk).
+
+<a name="anbox-streaming-sdk"></a>
+### Anbox Streaming SDK
+
+An SDK that allows the development of custom streaming clients, using JavaScript or C/C++.
+
+See [Anbox Streaming SDK](https://discourse.ubuntu.com/t/anbox-cloud-sdks/17844#streaming-sdk).
+
+<a name="android-app"></a>
+### Android app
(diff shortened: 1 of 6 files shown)


reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### AMS SDK
+
+An SDK that provides Go code for connecting to AMS through the exposed REST API.
+
+See [AMS SDK](https://discourse.ubuntu.com/t/anbox-cloud-sdks/17844#ams-sdk).
+
+<a name="anbox"></a>
+### Anbox
+
+An open-source project that provides a container-based approach to boot a full Android system on a regular GNU/Linux system. Anbox Cloud uses Anbox.
```
ru-fu, line 105: Can we add more information on how Anbox Cloud uses Anbox? I didn't find anything.

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### Container
+
+One of the main objects of Anbox Cloud. Every time you launch an application or image, Anbox Cloud creates a container for it. Every container provides a full Android system.
+
+See https://discourse.ubuntu.com/t/about-containers/17763.
+
+<a name="core-stack"></a>
+### Core stack
+
+The core parts of the Anbox Cloud stack that are required for all deployments.
```
ru-fu, line 244: Is this true that all parts of the core stack are required for all deployments? Can we add more information on what parts those are?

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### LXD
+
+A system container and virtual machine manager that offers a unified user experience around full Linux systems running inside containers or virtual machines. Anbox Cloud is based on LXD.
+
+See [the LXD website](https://linuxcontainers.org/).
+
+<a name="monitoring-stack"></a>
+### Monitoring stack
+
+The part of the Anbox Cloud stack that provides monitoring functionality.
```
ru-fu, line 312: If I understood it correctly, the monitoring stack is an optional part of the core stack? (Which contradicts the definition of the core stack ;) ).

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### Watchdog
+
+A software component that monitors the app in a container and terminates the container if the app crashes or is moved to the background.
+
+See [Watchdog settings](https://discourse.ubuntu.com/t/application-manifest/24197#watchdog).
+
+<a name="web-dashboard"></a>
+### Web dashboard
+
+A web GUI for Anbox Cloud from where developers can create, manage and stream applications from their web browser. Previously called Dev UI.
```
ru-fu, line 436: Is "previously called" okay, or do we need to state that it is a replacement?

reference/glossary.md
aguadoenzo: Do we want to specify this here ? The open source Anbox and Anbox Cloud are starting to diverge in quite a few ways now

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+
+<a name="boot-package"></a>
+### Boot package
+
+The package to launch in an application container once the system has booted.
+
+<a name="bootstrap-process"></a>
+### Bootstrap process
+
+The process that builds the application and synchronises it across all LXD nodes in the cluster when creating an application.
```
aguadoenzo, line 230:
```suggestion
The process that builds and optimises the application to run on Anbox Cloud
```
The bootstrap process doesn't technically sync an application on the whole cluster, LXD does that automatically when a new container is launched on a node (which is why the first container created for an application takes longer to boot)

reference/glossary.md
aguadoenzo: Correct. A bare minimum deployment of Anbox Cloud would require:
 - ams
 - etcd
 - at least 1 LXD worker
 - 1 node controller per LXD worker
 - easyrsa

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+
+<a name="gpu"></a>
+### Graphics Processing Unit (GPU)
+
+A specialised processor that is designed to accelerate image processing and graphics rendering for output to a display device.
+
+<a name="ha"></a>
+### High availability (HA)
+
+The characteristic of a system to continuously be available without failing for a higher-than-normal period of time.
```
aguadoenzo, line 270: While that definition is correct, I would be a bit more specific to what it means in our case here. In our case HA means avoiding having a single point of failure by having replicas of every service

reference/glossary.md
aguadoenzo: It's a recommended but indeed optional part of the stack. You can still run containers without it

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+
+<a name="raw-container"></a>
+### Raw container
+
+A container that is created when launching an image. It runs the full Android system, without any additional apps installed.
+
+See [Application containers vs. raw containers](https://discourse.ubuntu.com/t/about-containers/17763#application-vs-raw).
+
+<a name="regular-container"></a>
+### Regular container
```
aguadoenzo, line 352: Do we need to explain this ? I feel like it just brings more confusion, it's the same as the definition for `container`

reference/glossary.md
aguadoenzo: I'm even wondering if it's worth mentioning the Dev UI at all. It was phased out quite a few releases ago
Technically speaking though, it's a full blown out replacement, it's a different name, code, deployment method, and feature set

reference/glossary.md
ru-fu: I think it's important to point out that there is something called Anbox and that it is different from Anbox Cloud, and to do that, we need to explain how they differ.

reference/glossary.md
ru-fu: So the core stack contains all these services(?) ?

reference/glossary.md
ru-fu: Currently we are making a distinction between regular containers and base containers, both in the docs and in the container type. Can we get rid of that distinction?

reference/glossary.md
ru-fu: "Dev UI" still occurs in the docs, but maybe it would be better to get rid of it there than to mention it here?

reference/glossary.md
aguadoenzo: Correct

reference/glossary.md
aguadoenzo: I vote to get rid of it in the docs. I think it would mostly create confusion, what do you think @adglkh @morphis ?

reference/glossary.md
aguadoenzo: Not really, we have to make a distinction between a base container and a regular container. However in the context of the glossary, `container` and `regular container` are the same type of containers
Although it makes sense to specify both, considering `container` is the generic term and `regular container` is just there to contrast the definition of `base container`
After consideration, let's leave it as is

reference/glossary.md
morphis: Having an explanation here is great as this is a thing which is often misunderstood (and the naming doesn't make it any better). Reality is that Anbox and Anbox Cloud are completely independent things. They don't share nothing but the name "Anbox" and similar ideas. Technically Anbox Cloud is not using Anbox. We need to make this clear in here.

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+
+<a name="ams-node-controller"></a>
+### AMS Node Controller
+
+A service that runs on every LXD node and puts the appropriate firewall rules in place when a container is started or stopped.
+
+<a name="ams-sdk"></a>
+### AMS SDK
+
+An SDK that provides Go code for connecting to AMS through the exposed REST API.
```
morphis, line 98:
```suggestion
An SDK that provides Go language bindings for connecting to AMS through the exposed REST API.
```

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### Anbox
+
+An open-source project that provides a container-based approach to boot a full Android system on a regular GNU/Linux system. Anbox Cloud uses Anbox.
+
+See [the Anbox project on GitHub](https://github.com/anbox/anbox).
+
+<a name="aam"></a>
+### Anbox Application Manager (AAM)
+
+A utility (`aam`) that is installed in the Android image and that can be used to back up and restore application data.
```
morphis, line 112:
```suggestion
A utility (`aam`) that is installed in the Anbox image and that can be used for various thigns including back up and restore of Android application data.
```

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### Anbox Cloud Appliance
+
+A self-contained, single-machine deployment variant of Anbox Cloud.
+
+See [Variants](https://discourse.ubuntu.com/t/about-anbox-cloud/17802#variants).
+
+<a name="amc"></a>
+### Anbox Management Client (AMC)
+
+The command line interface that is used to manage the Anbox Management Service.
```
morphis, line 140:
```suggestion
The command line interface that is used to manage the Anbox Management Service (AMS).
```

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+
+<a name="ams"></a>
+### Anbox Management Service (AMS)
+
+The service that handles all aspects of the application and container life cycle in Anbox Cloud. AMS is responsible for managing containers, applications, addons, updates and more, ensuring high density, performance and fast container startup times.
```
morphis, line 145: I would add a sentence explaining that AMS is using etcd as database and is connecting to LXD over its REST API

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+
+See https://discourse.ubuntu.com/t/anbox-management-service-ams/24321.
+
+<a name="anbox-platform-sdk"></a>
+### Anbox Platform SDK
+
+An SDK that provides support for developing custom platform plugins, which allows cloud providers to integrate Anbox with their existing infrastructure.
```
morphis, line 152:
```suggestion
An C/C++ SDK that provides support for developing custom platform plugins, which allows users to integrate Anbox with their existing infrastructure.
```

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### Hook
+
+Code that is invoked at different points in time in the life cycle of a container. Hooks are part of addons.
+
+See [Hooks](https://discourse.ubuntu.com/t/addons/25293#hooks).
+
+<a name="image"></a>
+### Image
+
+The base for a container, which contains all necessary components like Anbox or the Android root file system. Anbox Cloud provides images based on different Android versions and different architectures.
```
morphis, line 284:
```suggestion
The base for a container, which contains all necessary components like Anbox or the Android root file system. Anbox Cloud provides images based on different Android and Ubuntu versions and different architectures.
```

reference/glossary.md
morphis: I would like to rephrase this here and make the "monitoring stack" a generic thing which can be implemented in different ways. We as Anbox Cloud team don't provide such a thing but can make recommendations what to use.

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### Software Rasterization (swrast)
+
+An LLVMpipe-based software rendering platform that is useful for visual tests but does not provide audio input/output.
+
+See https://discourse.ubuntu.com/t/anbox-platforms/18733.
+
+<a name="stream-agent"></a>
+### Stream agent
+
+The software running on a server equipped with a GPU connected to Anbox Cloud, which serves as an entry point that the stream gateway can connect to.
```
morphis, line 389: It doesn't necessarily need to be a server with a GPU. The stream agent connects AMS to the stream gateway and allows distribution from the gateway to multiple independent AMS installations.

morphis, line 389: It doesn't necessarily need to be a server with a GPU. The stream agent connects AMS to the stream gateway and allows distribution from the gateway to multiple independent AMS installations.

reference/glossary.md
```diff
@@ -0,0 +1,445 @@
...
+### Stream gateway
+
+The central component that connects clients with stream agents. Its role is to choose the best possible region depending on the user location and server capacities.
+
+See https://discourse.ubuntu.com/t/about-application-streaming/17769.
+
+<a name="streaming-stack"></a>
+### Streaming stack
+
+A collection of components designed to run containers on GPU-equipped machines and stream their visual output to clients via WebRTC.
```
morphis, line 403: Not necessarily GPU equipped machines. It can run on machines without GPUs too using software rendering

reference/glossary.md
morphis: Yes, lets get rid of "Dev UI" as term.

morphis (changes requested):

reference/glossary.md
ru-fu: Change to something like:
"A component of Anbox Cloud that facilitates booting an Android system on a regular GNU/Linux system. While the concepts of the Anbox component in Anbox Cloud are similar to the [Anbox open source project](https://github.com/anbox/anbox), the Anbox open source project is an independent project that is not related to or used in Anbox Cloud.

reference/glossary.md
```diff
@@ -0,0 +1,451 @@
...
+### AMS SDK
+
+An SDK that provides Go language bindings for connecting to AMS through the exposed REST API.
+
+See [AMS SDK](https://discourse.ubuntu.com/t/anbox-cloud-sdks/17844#ams-sdk).
+
+<a name="anbox"></a>
+### Anbox
+
+A component of Anbox Cloud that facilitates booting an Android system on a regular GNU/Linux system. The concepts of the Anbox component in Anbox Cloud are similar to the [Anbox open source project](https://github.com/anbox/anbox), but the Anbox open source project is an independent project that is not related to or used in Anbox Cloud.
```
ru-fu, line 105: Does this make sense?

reference/glossary.md
```diff
@@ -0,0 +1,451 @@
...
+### LXD
+
+A system container and virtual machine manager that offers a unified user experience around full Linux systems running inside containers or virtual machines. Anbox Cloud is based on LXD.
+
+See [the LXD website](https://linuxcontainers.org/).
+
+<a name="monitoring-stack"></a>
+### Monitoring stack
+
+A reference implementation for basic monitoring functionality based on Prometheus, Grafana and Telegraf. Anbox Cloud does not provide a full monitoring solution, but the reference implementation can be used as a starting point for implementing a custom solution.
```
ru-fu, line 318: Does this make sense?

adglkh:
As we mentioned `adb` in the glossary, I feel like it's worth describing the `anbox-shell` in this section too.
```
anbox-shell -> an ADB shell with root permission granted
```

reference/glossary.md
morphis: Perfect, thanks!

reference/glossary.md
morphis: Perfect, thanks!

morphis (approved):
LGTM, thanks!

aguadoenzo (approved):

PR #88, opened 2022-02-11
add documentation for the anbox-cloud-appliance tool

ru-fu:


ru-fu:
Not quite sure I'm happy with this addition - it exposes that we aren't clear about which parts of the documentation apply to Anbox Cloud and which to the Anbox Cloud Appliance ...

morphis (approved):
I agree, we need to make the distinction a bit clearer (rule of thumb, everything which says "juju" doesn't apply for the appliance and everything else does)

LGTM for a first shot to get this part documented

PR #91, opened 2022-02-15
Add release notes for 1.11.5

adglkh:


release-notes.md
```diff
@@ -139,6 +139,20 @@ None.
 
 [/Details]
 
+[Details=1.11.5]
```
ru-fu, line 142: Should this go here or at the top? 1.9.5 is above 1.10 because it was released later.

component-versions.md
```diff
@@ -2,6 +2,35 @@ This documents the versions of the different components for each Anbox Cloud rel
 
 Not all components are updated with each release. When components are not updated this is called out in the [release notes](https://discourse.ubuntu.com/t/release-notes/17842) and components are marked with `n/a` below.
 
+[Details=1.11.5]
```
ru-fu, line 5: This one should probably go between 1.11.4 and 1.12.0 - that makes it easier to find, and for the component version, the time when the release came out is less important than for the release notes.

ru-fu (approved):

PR #94, opened 2022-02-24
Document how to install APK as a system app

adglkh:


howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,70 @@
+Anbox Cloud allows user applications to be installed as Android system apps. This can be done by using the `aam install-system-app` command.
```
ru-fu, line 1: Can you add a short explanation on the difference between a system app and a normal app? Why would a user want an app to be installed as system app?

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Make a system app
+To be a system app running in the system process of Android system, you must add the attribute `android:sharedUserId="android.uid.system"` into the `<manifest>` or `<application>` tag in the AndroidManifest.xml file of your Android project.
```
ru-fu, line 4: Is it important whether you put it in the manifest tag or the application tag?
If it doesn't really matter, we should mention only one of them. If there is a difference, we need to explain it.

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,70 @@
...
+...
+```
+
+Then build and sign the application alongside with other Android applications when building your Android image or sign it with [Android Studio](https://developer.android.com/studio/publish/app-signing). For the latter approach, the system platform key is not required and you can use the keys that are generated from Android Studio to sign the application.
+
+### Create an addon
+
+By creating an addon and invoking the `aam install-system-app` command in a `pre-start` hook, the signed APK can be installed as a system app in the Android container.
+
+Following the [tutorial](https://discourse.ubuntu.com/t/creating-an-addon/25284) and create the basic layout of an addon, which has a manifest.yaml file and hooks folder prepared in advance. Assumed that the APK file is named `app.apk`, place the APK into the addon folder and create a `pre-start` hook with the following content.
```
ru-fu, line 21: Not sure I understand "which has a manifest.yaml file and hooks folder prepared in advance". Do you mean the user should prepare this as described in the tutorial?

ru-fu (commented):
A few content questions. I'll do a copy edit later.

The overall structure looks good to me. :)

Can you add this to the landing page as well?  (https://anbox-cloud.io/docs/howto/port/landing)
Not sure if it'll fit in the current list though. It might fit in if you'll encounter problems if you try to install a system app in the usual way - is that the case?

howto/port/install-as-system-app.md
adglkh: Yes. that's what a basic skeleton of an addon looks like.
We were thinking in the past to introduce a command like 
- `amc addon create`
- `amc application create` 
to provide an interactive way for end user on an application/addon creation. Just like `lxd init` does today
```
$ sudo lxd init
Name of the storage backend to use (dir or zfs) [default=zfs]: zfs
Create a new ZFS pool (yes/no) [default=yes]? no
Name of the existing ZFS pool or dataset: lxd
Would you like LXD to be available over the network (yes/no) [default=no]? no
Do you want to configure the LXD bridge (yes/no) [default=yes]? no
```
So after a user answered all questions, we have a manifest.yaml file well prepared with all requirements filled up to create an app/addon. In this way, the app/addon creation would be more intuitive than what it works today.
(I guess we still have it in our backlog.)

howto/port/install-as-system-app.md
adglkh: It matters in the way that those are only two tags that the attribute `android:sharedUserId` should be put into. Going through AOSP code base, looks like the attribute is put into `<manifest>` tag in the source tree of almost all the system apps. I think we should follow.

howto/port/install-as-system-app.md
adglkh: Will do

howto/port/install-as-system-app.md
adglkh: It matters in the way that those are only two tags that the attribute `android:sharedUserId` should be put into. Going through AOSP code base, looks like the attribute is put into `<manifest>` tag in the source tree of almost all the system apps. I think we should follow.

howto/port/install-as-system-app.md
adglkh: Will do

adglkh:
Can you add this to the landing page as well? (https://anbox-cloud.io/docs/howto/port/landing)
Not sure if it'll fit in the current list though. It might fit in if you'll encounter problems if you try to install a system app in the usual way - is that the case?
A: Ah, I missed that. Thanks. It's definitely worth mentioning the system app installation on that page.

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,75 @@
+Normally, a user app is signed by the developer and restricted to fewer permissions at runtime while a system app is usually [signed with the platform key](https://source.android.com/devices/tech/ota/sign_builds) when building an Android image, it's pre-installed under the system partition and running with some ["signature" protection level permissions](https://developer.android.com/guide/topics/manifest/permission-element.html#plevel) in the system process of Android container. An application must be running as a system app if it
```
morphis, line 1: The `in the system process` is not correct as that indicates that the app runs as part of the `system_server` process which it does not. It's more running as independent process but with system privileges.

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,75 @@
...
+
+- requires to access Android hidden APIs
+- gains some signature protection level permissions
+
+Anbox Cloud allows a user app to be installed as a system app in Android container. This can be done by using the `aam install-system-app` command.
+
+### Make a system app
+To be a system app running in the system process of Android system, you must add the attribute `android:sharedUserId="android.uid.system"` into the `<manifest>` tag in the AndroidManifest.xml file of your Android project.
```
morphis, line 9:
```suggestion
To be a system app running as a process with system privileges, you must add the attribute `android:sharedUserId="android.uid.system"` into the `<manifest>` tag in the AndroidManifest.xml file of your Android application.
```

morphis: We should also reference the snippet below

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,75 @@
...
+```xml
+...
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    package="<package_name>"
+    android:sharedUserId="android.uid.system">
+...
+```
+
+Then build and sign the application alongside with other Android applications when building your Android image or sign it with [Android Studio](https://developer.android.com/studio/publish/app-signing). For the latter approach, the system platform key is not required and you can use the keys that are generated from Android Studio to sign the application.
```
morphis, line 20:
```suggestion
Then build and sign the application alongside with other Android applications when building your Android image or sign it with [Android Studio](https://developer.android.com/studio/publish/app-signing). For the latter, the system platform key is not required and you can use the keys that are generated from Android Studio to sign the application.
```

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,75 @@
+Normally, a user app is signed by the developer and restricted to fewer permissions at runtime while a system app is usually [signed with the platform key](https://source.android.com/devices/tech/ota/sign_builds) when building an Android image, it's pre-installed under the system partition and running with some ["signature" protection level permissions](https://developer.android.com/guide/topics/manifest/permission-element.html#plevel) in the system process of Android container. An application must be running as a system app if it
+
+- requires to access Android hidden APIs
+- gains some signature protection level permissions
+
+Anbox Cloud allows a user app to be installed as a system app in Android container. This can be done by using the `aam install-system-app` command.
```
morphis, line 6: Shouldn't we give a quick summary here of the steps necessary we then described further below?

howto/port/install-as-system-app.md
morphis: I think we should show the full addon here rather than referencing another howto. That would make this self contained, ideas?

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,75 @@
...
+```bash
+#!/bin/bash -ex
+
+# Only install the APK as a system app when bootstrapping an application.
+if  [ "$CONTAINER_TYPE" = "regular" ]; then
+  exit 0
+fi
+
+aam install-system-app \
+  --apk="${ADDON_DIR}/app.apk \
```
morphis, line 37:
```suggestion
  --apk="${ADDON_DIR}"/app.apk \
```

morphis (changes requested):

howto/port/install-as-system-app.md
adglkh: +1

howto/port/install-as-system-app.md
adglkh: +1

howto/port/install-as-system-app.md
adglkh: Will fix. 
I think we should show the full addon here rather than referencing another howto. That would make this self contained, ideas?
A: I avoid doing that. The reason is that if we do that, we may replicate some other bits which existed in the tutorial section. For this page, I just want to highlight the stuff that people need to do to install apk as a system app without introducing some common knowledge/syntaxes that can be found from the tutorial which would be the first place people could get hands on the addon creation.

morphis: Ok, fine for me

ru-fu: I agree with Gary - we shouldn't duplicate content. And advanced users will know how to create an addon without further instructions anyway.

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,86 @@
...
+└── manifest.yaml
+```
+
+Add this addon to AMS with the following command:
+
+    amc addon add install-sys-app .
+
+### Include the addon in an application
+
+To use this addon in an application, include the addon name under `addons` in the application manifest file when creating an application. You must also enable the feature `allow_custom_system_signatures`, which ensures that the `aam install-sys-app` command that is invoked in the `pre-start` hook of the addon works properly.
```
adglkh, line 73: which ensures that the `aam install-sys-app` command
->
which ensures that the `aam install-system-app` command

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,86 @@
...
+.
+├── app.apk
+├── hooks
+│   └── pre-start
+└── manifest.yaml
+```
+
+Add this addon to AMS with the following command:
+
+    amc addon add install-sys-app .
```
adglkh, line 69: Let's use `install-system-app` as the addon name for clarity and consistent purpose.

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Include the addon in an application
+
+To use this addon in an application, include the addon name under `addons` in the application manifest file when creating an application. You must also enable the feature `allow_custom_system_signatures`, which ensures that the `aam install-sys-app` command that is invoked in the `pre-start` hook of the addon works properly.
+
+```yaml
+...
+addons: [ install-sys-app ]
```
adglkh, line 77: `install-sys-app`
->
`install-system-app`

adglkh (commented):
Thanks for this! I just found one bug in my previous writeup when reviewing this.

howto/port/install-as-system-app.md
```diff
@@ -0,0 +1,86 @@
...
+  exit 0
+fi
+
+aam install-system-app \
+  --apk="${ADDON_DIR}"/app.apk \
+  --permissions=<comma-separated list of permissions that the application requires> \
+  --package-name=<package_name>
+```
+
+The values of the `package_name` and the `permissions` parameters must match the ones defined in the `AndroidManifest.xml` file of the Android project. If the app requires access to hidden Android APIs to function, add the `--access-hidden-api` parameter to the above command. Use `aam install-system-app --help` for details about this command.
```
ru-fu, line 55: package-name, not package_name
(I'll fix this during publishing).

ru-fu (approved):
One small typo - but I'll fix that during publishing. :)

PR #123, opened 2022-05-10
Add jammy images to list of available ones

morphis:


adglkh (approved):
LGTM

ru-fu:
Should we also update this in examples where we currently load bionic images? Or should we leave that as a separate task?

ru-fu (approved):

PR #137, opened 2022-06-13
Add jammy support

ru-fu:
Once we have CI testing in place, recommend using Jammy for new installations.


ru-fu:
Updated the PR to contain only the Jammy support for the appliance.
The full changes (updating images in examples) are here: https://github.com/anbox-cloud/discourse-docs/compare/main...ru-fu:full-jammy-support?expand=1

adglkh (approved):
LGTM

PR #143, opened 2022-06-30
Document new /1.0/platform API endpoint for Anbox

morphis:


ref/anbox-https-api.md
```diff
@@ -445,3 +445,48 @@ Return value:
...
 As a result, a trace file can be found from the given path recorded in the response.
 You can pull that file from the container and import it to [Perfetto Trace Viewer](https://ui.perfetto.dev/#!/viewer) for further analysis.
+
+<a name="heading--10platform"></a>
+### `/1.0/platform`
+#### GET
+
+* Description: Get information about the current platform Anbox uses
+ * Operation: sync
+ * Return: Information about the current platform Anbox uses
```
ru-fu, line 455:
```suggestion
* Description: Get information about the current platform that Anbox uses
* Operation: sync
* Return: Information about the current Anbox platform
```

ref/anbox-https-api.md
```diff
@@ -445,3 +445,48 @@ Return value:
...
+  "status": "Success",
+  "status_code": 200,
+  "type": "sync"
+}
+```
+
+#### PATCH
+ * Description: Update configuration of the platform currently used by Anbox
+ * Operation: sync
+ * Return: standard return value or standard error
```
ru-fu, line 476:
```suggestion
 * Return: Standard return value or standard error
```

ref/anbox-https-api.md
```diff
@@ -445,3 +445,48 @@ Return value:
...
+
+```bash
+{
+ "status": "Success",
+ "status_code": 200,
+ "type": "sync"
+}
+```
+
+The available configuration items available depend on the platform being used by Anbox and are dynamically registered. The following table shows a list of items available with the platforms shipping with Anbox Cloud.
```
ru-fu, line 488:
```suggestion
The available configuration items depend on the platform being used by Anbox and are dynamically registered. The following table shows a list of items available with the platforms shipping with Anbox Cloud.
```

ref/anbox-https-api.md
```diff
@@ -445,3 +445,48 @@ Return value:
...
+
+Platform | Field name        | JSON type | Description        |
+---------|-------------------|-----------|-
+`webrtc` | `rtc_log`         | boolean   | Enable/disable [RTC event logging](https://webrtc.googlesource.com/src/+/lkgr/logging/g3doc/rtc_event_log.md). Logs are written to `/var/lib/anbox/traces/rtc_log.*` inside the Anbox container |
```
ru-fu, line 492:
```suggestion
`webrtc` | `rtc_log`         | Boolean   | Enable/disable [RTC event logging](https://webrtc.googlesource.com/src/+/lkgr/logging/g3doc/rtc_event_log.md). Logs are written to `/var/lib/anbox/traces/rtc_log.*` inside the Anbox container. |
```

ru-fu (changes requested):

ref/anbox-https-api.md
```diff
@@ -445,3 +445,48 @@ Return value:
...
+* Description: Get information about the current platform that Anbox uses
+* Operation: sync
+* Return: Information about the current Anbox platform
+
+Return value for `curl -s -X GET --unix-socket /run/user/1000/anbox/sockets/api.unix s/1.0/platform | jq .`:
+
+```bash
+{
+  "metadata": {
+    "name": "webrtc",
```
adglkh, line 462: When I read this, it might create a bit confusion after the webrtc platform now is the unified platform for all platforms(null/swrast/webrtc) that we had supported before in LXD image.
E.g. launch a null platform with
```
amc launch -p null -r
```
then issue the above curl command inside of the Anbox container, people will get the above response. Ideally, the platform name should be `null` rather than `webrtc`. Furthermore, if the `null` platform is specified, we should not list  config items like `"config":{"rtc_log":true}` but leave it blank.
For now, I don't have a good idea how to deal with this situation, and you? If not, I'm fine to leave it as is.

morphis: Even with `null` we can do webrtc nowadays as we recently added support for it. So `null` is has now become more of a configuration detail of the `webrtc` platform. Generally I think the name `webrtc` is wrong as well but I don't have any better idea. We started to migrate from `null` to `--enable-graphics` / `--disable-graphics` in the past and I think that's the right way forward.

adglkh (approved):
LGTM

ru-fu (approved):

ru-fu (approved):

PR #146, opened 2022-07-11
Drop tracing feature flag as no longer available

morphis:


ru-fu (approved):
This should probably be mentioned as removed in the release notes?

PR #155, opened 2022-08-15
correct the declaration location for `screen` settings

adglkh:
The key `Screen` should be part of AnboxStreamGatewayConnector
settings rather than AnboxStream settings.

### DIFF
--- a/howto/stream/web-client.md
+++ b/howto/stream/web-client.md
@@ -19,16 +19,16 @@ Detailed explanations can be found below.
         session: {
             app: "com.foo.bar",
         },
+        screen: {
+            width: 1280,
+            height: 720,
+            fps: 25,
+        }
     });
 
     let stream = new AnboxStream({
         connector: connector,
         targetElement: "anbox-stream",
-        screen: {
-            width: 1280,
-            height: 720,
-            fps: 25,
-        },
         controls: {
             keyboard: true
         },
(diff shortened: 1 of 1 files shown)


ru-fu (approved):

ru-fu (changes requested):
This occurs several times in this page - please check further down as well.

There isn't much explanation about the "screen" parameter, but the sentence "The default behaviour of the video is to fill the maximum space given by this element while keeping aspect ratio intact." should probably also be moved up to the section where the parameter is set.

adglkh:
Thanks for spotting on. 

I will drop all remaining `screen` settings from the AnboxStream object as they are not supposed to be there.
Regarding
```
"The default behaviour of the video is to fill the maximum space given by this element while keeping aspect ratio intact." should probably also be moved up to the section where the parameter is set.
```
This sentence is going to explain the style parameters for the video container element `anbox-stream`  
```
`<div id="anbox-stream" style="width: 100vw; height: 100vh;"></div>`
```
 we should keep it here for the context .

howto/stream/web-client.md
```diff
@@ -67,16 +72,12 @@ In this case, the connector is made to talk directly to the Stream Gateway and t
 `gateway.url.net` should be replaced with the Stream Gateway IP/Domain name. You can get this information by running `juju status`.
 `YOUR_AUTH_TOKEN` is a token [generated with the Stream Gateway](https://discourse.ubuntu.com/t/managing-stream-gateway-access/17784).
 `com.foo.bar` is the name of an application added via `ams`. You can list all applications by running `amc list`.
+`screen` setting allows you to define the display resolution and frame rate for the Android container.
```
ru-fu, line 75:
```suggestion
The `screen` setting allows you to define the display resolution and frame rate for the Android container.
```

PR #171, opened 2022-10-06
add "How to debug streaming issues"

ru-fu:


howto/stream/debug.md
```diff
@@ -0,0 +1,36 @@
+Debugging issues that occur when streaming your application can be tricky. The following instructions give you some pointers on how to start tracking down streaming issues.
+
+First of all, check the error message that occurs. Unfortunately, it is often hard to determine why a connection fails, which is why the error message does not always give a clear indication of the source of the error. Typical errors are covered though, so in these cases, the error message should give you an idea on where to look.
```
morphis, line 3: I think we can do a lot better in the SDK to generate more valuable error messages. Going to plan work for this.

howto/stream/debug.md
```diff
@@ -0,0 +1,36 @@
...
+
+If the error message does not help, check the WebRTC debug information. You can find this information in your browser:
```
morphis, line 5: We currently only support browser based clients so I think assuming the client is always browser based here is fine.

morphis (approved):
Thanks, LGTM!

howto/stream/debug.md
```diff
@@ -0,0 +1,36 @@
...
+
+You can create a dump of the debug information by expanding the **Create Dump** section (in Chrome) or clicking **Save Page** (in Firefox).
+
+Analyse the dump to find the source of the error, or provide it to Anbox Cloud's support team for help.
+
+## Streaming does not start
+
+If streaming does not start, there is usually a connection problem.
+
+If you are using the web dashboard and, for example, creating an application works as expected, the connection problem is not that the web dashboard cannot reach Anbox Cloud. Most likely, there are some security rules in place that prevent WebRTC from establishing a connection. In this case, check your firewall configuration.
```
adglkh, line 18: +1, firewall configuration and VPN setup.

howto/stream/debug.md
adglkh: From the developer's perspective, they can use `experimental.debug=false` option which would yield out debug information and help to diagnose session info as well E.g. sdp info. 
(I'm not quite sure if it's the right place to put it here.)

ru-fu: @adglkh Where would you set this option?

adglkh: With the following example.
https://github.com/anbox-cloud/discourse-docs/blob/2764c5620f5350052db99b968f0a784a7aa660be/howto/stream/web-client.md
People can enable the debug logging through

```
    let stream = new AnboxStream({
        connector: connector,
        targetElement: "anbox-stream",
        controls: {
            keyboard: true
        },
+       experimental: {
+           debug: true,
+       },
        callbacks: {
            error: error => {
                console.log("AnboxStream failed: ", error);
            }
        }
    });
    stream.connect();
    </script>
```

adglkh: As of now, it's impossible. We had a quick discussion about this just a little moment ago, we're going to support this in the dashboard, this task will be prioritized by @edlerd and @morphis . As an outcome, people can enable it in the stream view.

ru-fu: Thanks, I've added the information about experimental.debug to the web-client page.

adglkh:
LGTM

PR #180, opened 2022-11-10
feat(ams): document container.apt_mirror config option

morphis:


ref/ams-configuration.yaml
```diff
@@ -37,6 +37,13 @@ all:
       Maximum number of published versions per application. If the
       number of versions of an application exceeds this configuration,
       AMS will automatically clean up older versions.
+  container.apt_mirror:
+    type: "string"
+    default: "-"
+    description: >-
+      APT mirror to use within the containers. By default
```
ru-fu, line 44:
```suggestion
      APT mirror to use within the containers. By default,
```

ref/ams-configuration.yaml
```diff
@@ -37,6 +37,13 @@ all:
...
+      `http://archive.ubuntu.com` (amd64) or `http://ports.ubuntu.com`
+      (arm64) is being used
```
ru-fu, line 46:
```suggestion
      (arm64) is used.
```

ru-fu:
APT will probably fail the spellcheck, but we can fix that afterwards (if it happens ;) ).

ru-fu (approved):
Is that for 1.16 or can it go out right away?

PR #198, opened 2023-01-18
clarify that the size limit is in Google Play, not Anbox Cloud

ru-fu:


morphis (approved):
Thanks, LGTM!

PR #201, opened 2023-01-27
add a reference for exposed network ports

ru-fu:

### DIFF
--- a/ref/network-ports.md
+++ b/ref/network-ports.md
@@ -1,0 +1,36 @@
+The following tables list the network ports that Anbox Cloud exposes.
+
+For the regular Anbox Cloud deployment, these port are used for communication between components, and to allow accessing the Anbox Cloud interface. For the Anbox Cloud Appliance, all components are installed on the same machine. Therefore, there is no need to expose ports for internal communication.
+
+<a name="anbox-cloud"></a>
+## Anbox Cloud
+
+| Service               | Port(s)     | Protocol  | Exposed externally | Required                     | Description                           |
+|-----------------------|-------------|-----------|--------------------|------------------------------|---------------------------------------|
+| AMS                   | 8444        | TCP       | no                 | yes                          | HTTPS API                             |
+| AMS                   | 20002       | TCP       | no                 | no                           | HTTPS Prometheus endpoint             |
+| AMS node controller   | 10000-11000 | UDP & TCP | yes                | no                           | Container service ports               |
+| Anbox Cloud Dashboard | 5000        | TCP       | no                 | no                           | HTTPS website                         |
+| Anbox Stream Agent    | 443         | TCP       | no                 | yes                          | HTTPS API                             |
+| Anbox Stream Gateway  | 4000        | TCP       | no                 | yes                          | HTTPS API                             |
+| Anbox Stream Gateway  | 7000        | TCP       | no                 | yes                          | Dqlite HA and API                     |
+| Coturn                | 5349        | TCP       | yes                | yes                          | STUN/TURN                             |
+| Coturn                | 5349        | UDP       | yes                | yes                          | STUN/TURN                             |
+| Coturn                | 50000-51000 | UDP       | yes                | no (unless TURN is required) | TURN relay ports                      |
+| etcd                  | 2379        | TCP       | no                 | yes                          | gRPC API                              |
+| HAProxy               | 80          | TCP       | yes                | no                           | HTTP (redirects to HTTPS on port 443) |
+| HAProxy               | 443         | TCP       | yes                | no                           | redirects to HTTPS website            |
+| LXD                   | 8443        | TCP       | no                 | yes                          | HTTPS API                             |
+| NATS                  | 4222        | TCP       | no                 | yes                          | NATS API                              |
+
+
+<a name="appliance"></a>
+## Anbox Cloud Appliance
+
+| Service             | Port(s)     | Protocol  | Required | Description                            |
+|---------------------|-------------|-----------|----------|----------------------------------------|
+| AMS node controller | 10000-11000 | UDP & TCP | no       | Container service ports                |
+| Coturn              | 5349        | UDP       | yes      | STUN/TURN                              |
+| Coturn              | 60000-60100 | UDP       | yes      | TURN relay ports                       |
+| Traefik             | 80          | TCP       | yes      | HTTP (redirects to HTTPS on port 443)  |
+| Traefik             | 443         | TCP       | yes      | redirects to the Anbox Cloud dashboard |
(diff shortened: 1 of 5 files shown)


ref/network-ports.md
```diff
@@ -0,0 +1,35 @@
+The following tables list the network ports that Anbox Cloud exposes.
+
+For the regular Anbox Cloud deployment, these port are used for communication between components, and to allow accessing the Anbox Cloud interface. For the Anbox Cloud appliance, all components are installed on the same machine. Therefore, there is no need to expose ports for internal communication.
+
+<a name="anbox-cloud"></a>
+## Anbox Cloud
+
+| Service               | Port(s)     | Protocol | Exposed externally | Required                     | Description               |
+|-----------------------|-------------|----------|--------------------|------------------------------|---------------------------|
+| AMS                   | 8444        | TCP      | no                 | yes                          | HTTPS API                 |
```
ru-fu, line 10: I changed this (and all other "HTTP") to "HTTPS" - is that okay or is it misleading in any way?

ref/network-ports.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+<a name="appliance"></a>
+## Anbox Cloud Appliance
+
+| Service               | Port(s)     | Protocol | Required | Description                           |
+|-----------------------|-------------|----------|----------|---------------------------------------|
+| AMS node controller   | 10000-11000 | UDP      | no       | Container service ports               |
+| AMS node controller   | 10000-11000 | TCP      | no       | Container service ports               |
+| Anbox Cloud Dashboard | 80          | TCP      | yes      | HTTP (redirects to HTTPS on port 443) |
```
ru-fu, line 32: Should we have that line for the full Anbox Cloud as well?

ref/network-ports.md
```diff
@@ -0,0 +1,35 @@
...
+| Anbox Stream Agent    | 443         | TCP      | no                 | yes                          | HTTPS API                 |
+| Anbox Stream Gateway  | 4000        | TCP      | no                 | yes                          | HTTPS API                 |
+| Anbox Stream Gateway  | 7000        | TCP      | no                 | yes                          | Dqlite HA and API         |
+| Coturn                | 5349        | TCP      | yes                | yes                          | STUN/TURN                 |
+| Coturn                | 5349        | UDP      | yes                | yes                          | STUN/TURN                 |
+| Coturn                | 50000-51000 | UDP      | yes                | no (unless TURN is required) | TURN relay ports          |
+| etcd                  | 2379        | TCP      | no                 | yes                          | gRPC API                  |
+| LXD                   | 8443        | TCP      | no                 | yes                          | HTTPS API                 |
+| NATS                  | 4222        | TCP      | no                 | yes                          | API                       |
```
ru-fu, line 22: Is that just API (it originally said "NATS API", but that is duplication), or HTTPS API or something?

ref/network-ports.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+<a name="appliance"></a>
+## Anbox Cloud Appliance
+
+| Service               | Port(s)     | Protocol | Required | Description                           |
+|-----------------------|-------------|----------|----------|---------------------------------------|
+| AMS node controller   | 10000-11000 | UDP      | no       | Container service ports               |
+| AMS node controller   | 10000-11000 | TCP      | no       | Container service ports               |
```
ru-fu, line 31: Here we have the same ports for UDP and TCP, while the full deployment has only TCP. Is it correct?

ref/network-ports.md
```diff
@@ -0,0 +1,35 @@
...
+| Anbox Cloud Dashboard | 80          | TCP      | yes      | HTTP (redirects to HTTPS on port 443) |
+| Anbox Cloud Dashboard | 443         | TCP      | yes      | HTTPS website                         |
+| Coturn                | 5349        | UDP      | yes      | STUN/TURN                             |
+| Coturn                | 60000-60100 | UDP      | yes      | TURN relay ports                      |
```
ru-fu, line 35: These ports are different than the ones for the full deployment - correct?

ref/network-ports.md
morphis: Makes sense

ref/network-ports.md
morphis: It's not a HTTP based protocol. NATS uses TLS secured TCP with text messages to allow communication between clients and the server. See https://docs.nats.io/reference/reference-protocols/nats-protocol (maybe we link this from here)?

And maybe we add here too that when deployed in Anbox Cloud we enable TLS and authentication. See https://docs.nats.io/nats-concepts/security

ref/network-ports.md
morphis: It's both UDP and TCP for both deployment types.

ref/network-ports.md
morphis: Yes, however in case of the appliance we have traefik running on the host which provides outside access to 80/443 and then routes as necessary to internal services. However the dashboard internally only listens on 5000 by default (HTTPS)

ref/network-ports.md
morphis: Yes, we expose less ports than with the full deployment due to limitations in NFT rules LXD generates (1000 ports are too much for it)

ref/network-ports.md
```diff
@@ -0,0 +1,35 @@
+The following tables list the network ports that Anbox Cloud exposes.
+
+For the regular Anbox Cloud deployment, these port are used for communication between components, and to allow accessing the Anbox Cloud interface. For the Anbox Cloud appliance, all components are installed on the same machine. Therefore, there is no need to expose ports for internal communication.
```
morphis, line 3: Shouldn't be "Anbox Cloud Appliance"?

ref/network-ports.md
ru-fu: OK, then NATS API actually makes sense.
I don't think we need additional links here - but we should add them in the architecture overview. And the security part we actually have already in https://anbox-cloud.io/docs/exp/security

ref/network-ports.md
ru-fu: Should we list 5000 then as well?
And if we don't have traefik for the full deployment, do we have the same kind of redirect?

ref/network-ports.md
ru-fu: Good catch, thanks!

ref/network-ports.md
morphis: Yes, on the full deployment we use HAProxy as a reverse proxy. 

Lets put the dashboard onto port 5000 in the table and mark it as HTTPS only. We can then add HAProxy for the full deployment covering 80/44 (where 80 redirects to 443) and traefik to the appliance with the same set of ports (and redirect)

ru-fu: Thanks - updated now. Please check again. :)

morphis (approved):

PR #213, opened 2023-02-24
Improve upgrade instructions

morphis:


howto/update/upgrade-anbox.md
```diff
@@ -6,13 +6,38 @@ The upgrade instructions detail the revisions each charm needs to be upgraded to
...
+
+You should also make sure:
+
+* Your deployment is running normally
+* Your Juju client and controller/models are running the latest version
+* You read the release notes for the version you are upgrading to, which will alert you to any important changes to the operation of your cluster
+
+[note type="information" status="Note"]
+
+The following assume you're using Juju >= 3.1. If you're using Juju 2.9 you have to map the following commands:
```
ajanon, line 21: So... Which should they use? Just above, you wrote: 
> Your Juju client and controller/models are running the latest version

Shouldn't that ensure they are running the latest version?

howto/update/upgrade-anbox.md
```diff
@@ -37,39 +62,87 @@ The deployed Juju charms need to be upgraded next.
...
 
-Run the following commands in the exact same order as listed here but skip those you don't use in your deployment:
+For any of the charm upgrades you can watch the upgrade status by running
+
+    juju status
+
+Only continue with the next step once the current step has completed successfully and all units in the output are marked as **active**.
+
+[note type="information" status="Note"]
+If you don't run Anbox Cloud in a high availity configuration, upgrading the charms will cause a short down time of individual service components during the process.
```
ajanon, line 72:
```suggestion
If you don't run Anbox Cloud in a high availability configuration, upgrading the charms will cause a short down time of individual service components during the process.
```

howto/update/upgrade-anbox.md
```diff
@@ -37,39 +62,87 @@ The deployed Juju charms need to be upgraded next.
...
+As first step we will update all infrastucture components. This includes deployed internal certificate authorities and etcd.
+
+First we update easyrsa:
+
+    juju refresh internal-ca --revision=26
+    juju refresh etcd-ca --revision=26
+
+Once the upgrade has completed, you can upgrade the etcd service next:
+
+    juju refresh etcd --revision=724
```
ajanon, line 86: Do we really need to specify the revision number? Isn't there a shortcut or alias?
Otherwise, this means we have to make sure we update the revision for every release.

howto/update/upgrade-anbox.md
```diff
@@ -37,39 +62,87 @@ The deployed Juju charms need to be upgraded next.
...
+To upgrade the registry, run
+
+    juju refresh --channel=1.17/stable aar
+
+### Upgrade control plane
+
+If you have the streaming stack deployed, you need to update the charms responsible for the control plane next. If you do not use the streaming stack, you can skip this step.
+
+[note type="information" status="Note"]
+If you don't run any of the services in a high availity configuration, upgrading the charms will cause a short down time of the service.
```
ajanon, line 101:
```suggestion
If you don't run any of the services in a high availability configuration, upgrading the charms will cause a short down time of the service.
```

howto/update/upgrade-anbox.md
```diff
@@ -37,39 +62,87 @@ The deployed Juju charms need to be upgraded next.
...
-    juju config ams-node-controller ua_token=<your token>
-    juju config aar ua_token=<your token>
-    juju config anbox-stream-gateway ua_token=<your token>
-    juju config anbox-stream-agent ua_token=<your token>
-    juju config anbox-cloud-dashboard ua_token=<your token>
+In case you want to upgrade individual LXD nodes separately, you can set the following configuration option for the LXD charm before running the refresh command above:
+
+    juju config lxd enable_manual_upgrade=true
+
+This will allow you to run the actual upgrade process for each deployed LXD instance separately. After the charm has been refreshed, the upgrade for a single LXD deployment unit can be triggered by running
```
keirthana, line 134:
```suggestion
This will allow you to run the actual upgrade process for each deployed LXD instance separately. After the charm has been refreshed, the upgrade for a single LXD deployment unit can be triggered by running:
```

howto/update/upgrade-anbox.md
```diff
@@ -37,39 +62,87 @@ The deployed Juju charms need to be upgraded next.
...
+
+    juju run --wait=30m lxd/0 upgrade
+
+Once the upgrade has completed the unit will be marked as active.
+
+For major and minor version upgrades, an update of the LXD charm may upgrade kernel modules or GPU drivers. This requiers stopping any running containers before applying the upgrade and performing a reboot of the machine once the upgrade completed.
+
+In case a reboot of the machine is required a status message will be shown. When the machine has been rebooted, the status message can be cleared by running
+
+    juju run --wait=1m lxd/0 clear-notification
```
keirthana, line 144:
```suggestion
Once the upgrade has completed, the unit will be marked as active.

For major and minor version upgrades, an update of the LXD charm may upgrade kernel modules or GPU drivers. This requires stopping any running containers before applying the upgrade and performing a reboot of the machine once the upgrade completed.

In case a reboot of the machine is required, a status message will be shown. When the machine has been rebooted, the status message can be cleared by running:

    juju run --wait=1m lxd/0 clear-notification
```

keirthana (changes requested):

howto/update/upgrade-anbox.md
```diff
@@ -6,13 +6,38 @@ The upgrade instructions detail the revisions each charm needs to be upgraded to
...
+| `juju exec` | `juju run` |
+
+[/note]
+
 ## Upgrade OS
 
 Before you run the upgrade of the charms, you should make sure all packages on the machines that are part of the deployment are up-to-date. To do so, run the following commands on each machine:
 
     sudo apt update
     sudo apt upgrade
```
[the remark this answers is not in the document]
ajanon: If you want to check how it looks like when rendered, you can check that file on the branch from the PR: https://github.com/morphis/discourse-docs/blob/improve-upgrade/howto/update/upgrade-anbox.md

Also, markdown renders indented text such as this one as monospace by default.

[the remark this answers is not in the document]
keirthana: Thanks, I didn't know that :) @morphis please ignore my comments about the code formatting, that's my bad.

howto/update/upgrade-anbox.md
morphis: We pin particular versions of the charm with each Anbox Cloud release as that is what we have tested with and qualified. So yes, we have to specify the exact revision here.

howto/update/upgrade-anbox.md
```diff
@@ -37,39 +62,87 @@ The deployed Juju charms need to be upgraded next.
...
+[note type="information" status="Note"]
+If you don't run any of the services in a high availity configuration, upgrading the charms will cause a short down time of the service.
+[/note]
+
+To upgrade all charms, run the following commands:
+
+    juju refresh --channel=1.17/stable anbox-cloud-dashboard
+    juju refresh --channel=1.17/stable anbox-stream-gateway
+    juju refresh --channel=1.17/stable anbox-stream-agent
+    juju refresh --channel=1.17/stable coturn
```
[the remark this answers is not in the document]
morphis: No, `latest/stable` is nothing we use for the charms. We change the default track to `<version>` and want users to stay on version specific tracks unless they explicitly want to upgrade.

howto/update/upgrade-anbox.md
morphis: It's tricky and depends on various other things (e.g. 3.x doesn't support bionic but Landscape still requires bionic so cannot be deployed with anything other but 2.9. Then not everyone is using Landscape, ...). 

From the upgrade perspective we just have to account for different versions.

keirthana (approved):
LGTM!

PR #217, opened 2023-03-13
Fix self-signed certificate wording

keirthana:
Fix the wording that is misleading in the procedure to add a self-signed certificate


morphis (approved):
LGTM, thanks!

PR #221, opened 2023-03-16
Fix messaging around use of LXD/Anbox images

keirthana:
Currently, we are using the terms LXD containers and images synonymous with Anbox containers and images. The idea is to be aware that within the context of Anbox Cloud, like when referring to activities related to AMS, they can be called Anbox images but when describing the overall architecture, it is better to explain from a big picture point of view that they are LXD images that contain Anbox configuration.


adglkh (approved):
LGTM

PR #222, opened 2023-04-03
Create rendering architecture topic

keirthana:
The images used are placeholders, we will get final images from the web/design team later to replace these images.

### DIFF
--- a/exp/rendering-architecture.md
+++ b/exp/rendering-architecture.md
@@ -35,7 +35,7 @@ For communication between the hardware composer module on the Android side and A
 
 ![Anbox Cloud NVIDIA pipeline|690x440](https://assets.ubuntu.com/v1/3ba1fddd-NVIDIA_pipeline.png)
 
-This NVIDIA pipeline uses the EmuGL stack. EmuGL is the hardware GLES emulation that is implemented with a mix of translator libraries, system libraries, and other rendering components. To understand more about EmuGL, see [EmuGL design overview](https://android.googlesource.com/platform/external/qemu/+/refs/heads/emu-master-dev/android/android-emugl/DESIGN). 
+This NVIDIA pipeline uses the EmuGL stack. EmuGL is the hardware GLES emulation that is implemented with a mix of translator libraries, system libraries, and other rendering components. To understand more about EmuGL, see [EmuGL design overview](https://android.googlesource.com/platform/external/qemu/+/6654d90be28d0059baf854caccd4caf9f7033ccb/android/android-emugl/DESIGN). 
 
 For NVIDIA, we cannot use the NVIDIA driver inside the Android container because of compatibility issues. Hence, we use the Enterprise Ready NVIDIA driver that is available on every Ubuntu installation. So by using the EmuGL stack, we have an Anbox Cloud GPU driver which is a standard OpenGL ES or EGL driver that receives the API calls and converts them to remote procedure calls to the NVIDIA driver. The actual rendering and actions on the NVIDIA driver happens on the Anbox runtime side inside the Ubuntu container and not in the Android space. 
 
--- a/exp/rendering-architecture.md
+++ b/exp/rendering-architecture.md
@@ -1,0 +1,52 @@
+# Supported GPU vendors
+
+Being a cloud solution, Anbox Cloud is optimised for GPUs that are designed for a data center. We currently support the following GPU vendors:
+
+* NVIDIA
+* Intel
+* AMD
+
+Anbox Cloud is extensively tested using NVIDIA GPUs and when needed, on Intel and AMD GPUs. However, if you want to use a different GPU vendor, you can customise and configure Anbox Cloud for the GPU vendor of your choice.
+
+# Supported GPU drivers
+
+For NVIDIA GPUs, Anbox Cloud uses the [Enterprise Ready Driver (ERD) from NVIDIA](https://help.ubuntu.com/community/NvidiaDriversInstallation) for Linux as available in Ubuntu. 
+For AMD and Intel GPUs, Anbox Cloud uses the [Mesa driver](https://www.mesa3d.org/).
+
+See [Component versions](https://anbox-cloud.io/docs/component-versions) to refer to the actual version supported for any particular Anbox Cloud release.
+
+# Supported APIs
+
+For NVIDIA GPUs Anbox Cloud uses OpenGL ES 3.2 and EGL graphical interfaces. For Intel and AMD GPUs, Anbox Cloud uses Vulkan 1.3. Both OpenGL and EGL are high level APIs while Vulkan is a lower level API that allows a better, granular level control over certain aspects of GPU management.
+
+# Rendering architecture
+
+For NVIDIA GPUs, Anbox Cloud driver stacks currently have a native OpenGL driver. For AMD and Intel GPUs, an OpenGL or EGL driver is layered on top of the Vulkan API. Since Vulkan API provides better GPU management at a lower level than OpenGL or EGL, this approach is beneficial and preferred by many users.
+
+To have a better understanding of the rendering architecture of Anbox Cloud, it is important to understand what the Android framework offers in terms of rendering. In Android, applications interact with the SurfaceFlinger which is the system compositor that is responsible for composing a frame together from all the outputs rendered by different applications. The frame is then submitted to the hardware composer which renders the frames on a screen. For more information on Android graphics components and how they work, see https://source.android.com/docs/core/graphics. 
+
+# Rendering pipeline
+
+Anbox Cloud has two rendering pipeline models - one for NVIDIA and the other for Intel and AMD. However, irrespective of the GPU that you use, the path of a frame typically looks like this: Android application > SurfaceFlinger > Hardware composer > Anbox Cloud > Display on screen or send it to streaming component.
+
+For communication between the hardware composer module on the Android side and Anbox runtime, we use [Wayland](https://wayland.freedesktop.org/). So Anbox Cloud really functions as a compositor for Android i.e., the hardware composer module receives frames from the SurfaceFlinger and submits them to the Anbox runtime using Wayland. The Anbox runtime then submits the frame towards its output, which is either the screen or the streaming component.
+
+## For NVIDIA
+
+![Anbox Cloud NVIDIA pipeline|690x440](https://assets.ubuntu.com/v1/3ba1fddd-NVIDIA_pipeline.png)
+
+This NVIDIA pipeline uses the EmuGL stack. EmuGL is the hardware GLES emulation that is implemented with a mix of translator libraries, system libraries, and other rendering components. To understand more about EmuGL, see [EmuGL design overview](https://android.googlesource.com/platform/external/qemu/+/refs/heads/emu-master-dev/android/android-emugl/DESIGN). 
+
+For NVIDIA, we cannot use the NVIDIA driver inside the Android container because of compatibility issues. Hence, we use the ERD NVIDIA driver that is available on every Ubuntu installation. So by using the EmuGL stack, we have an Anbox Cloud GPU driver which is a standard OpenGL ES or EGL driver that receives the API calls and converts them to remote procedure calls to the NVIDIA driver. The actual rendering and actions on the NVIDIA driver happens on the Anbox runtime side inside the ubuntu container and not in the Android space. 
+
+## For Intel and AMD
+
+![Anbox Cloud Intel and AMD pipeline|690x440](https://assets.ubuntu.com/v1/7aa9aff1-Intel_AMD_pipeline.png)
+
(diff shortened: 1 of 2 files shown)


keirthana:
The inclusive naming check fails because of a URL that contains 'master' in its path. I will have to ignore it because it is an external URL that we are just referring to and we don't have any control on it.

ajanon:
> The inclusive naming check fails because of a URL that contains 'master' in its path. I will have to ignore it because it is an external URL that we are just referring to and we don't have any control on it.

So, actually, instead of linking to the current master branch, you could link to this document for a specific commit:
https://android.googlesource.com/platform/external/qemu/+/6654d90be28d0059baf854caccd4caf9f7033ccb/android/android-emugl/DESIGN

If you find the link too long, you can also shorten the commit hash: https://android.googlesource.com/platform/external/qemu/+/6654d90b/android/android-emugl/DESIGN

8 characters for the hash should be enough to avoid collisions (but please don't quote me on that)

This has the benefit of being immutable (or it should be). Referencing the dynamic head of a branch may lead to problems in the future: if this file is removed or moved in the `emu-master-dev` branch, the link will become invalid.

In the future, if you wand to find this kind of link:
1. Go to the file you want a link for: https://android.googlesource.com/platform/external/qemu/+/refs/heads/emu-master-dev/android/android-emugl/DESIGN
1. Click on the [`[log]`](https://android.googlesource.com/platform/external/qemu/+log/refs/heads/emu-master-dev/android/android-emugl/DESIGN) link
1. Find the [most recent commit](https://android.googlesource.com/platform/external/qemu/+/6654d90be28d0059baf854caccd4caf9f7033ccb)
1. Choose the file you want to see in the list (`DESIGN` here)

exp/rendering-architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+# Rendering pipeline
+
+Anbox Cloud has two rendering pipeline models - one for NVIDIA and the other for Intel and AMD. However, irrespective of the GPU that you use, the path of a frame typically looks like this: Android application > SurfaceFlinger > Hardware composer > Anbox Cloud > Display on screen or send it to streaming component.
+
+For communication between the hardware composer module on the Android side and Anbox runtime, we use [Wayland](https://wayland.freedesktop.org/). So Anbox Cloud really functions as a compositor for Android i.e., the hardware composer module receives frames from the SurfaceFlinger and submits them to the Anbox runtime using Wayland. The Anbox runtime then submits the frame towards its output, which is either the screen or the streaming component.
```
adglkh, line 32: the hardware composer module receives frames from the SurfaceFlinger and submits them to the Anbox runtime using Wayland.
->
the hardware composer module receives frames from the SurfaceFlinger and notifies Anbox runtime using Wayland protocol.

The latter one is probably better as the Anbox doesn't receive the frames from the hw directly but reads them from the file descriptor.
Maybe @morphis come up with a better one

exp/rendering-architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+## For NVIDIA
+
+![Anbox Cloud NVIDIA pipeline|690x440](https://assets.ubuntu.com/v1/3ba1fddd-NVIDIA_pipeline.png)
+
+This NVIDIA pipeline uses the EmuGL stack. EmuGL is the hardware GLES emulation that is implemented with a mix of translator libraries, system libraries, and other rendering components. To understand more about EmuGL, see [EmuGL design overview](https://android.googlesource.com/platform/external/qemu/+/6654d90be28d0059baf854caccd4caf9f7033ccb/android/android-emugl/DESIGN). 
+
+For NVIDIA, we cannot use the NVIDIA driver inside the Android container because of compatibility issues. Hence, we use the Enterprise Ready NVIDIA driver that is available on every Ubuntu installation. So by using the EmuGL stack, we have an Anbox Cloud GPU driver which is a standard OpenGL ES or EGL driver that receives the API calls and converts them to remote procedure calls to the NVIDIA driver. The actual rendering and actions on the NVIDIA driver happens on the Anbox runtime side inside the Ubuntu container and not in the Android space.
```
adglkh, line 40: Probably add the following to the section:
```
In terms of performance, compared to the way that uses Vulkan as API on top of Angle when rendering on Intel or AMD GPUs, using the EmuGL stack is suboptimal as it's inefficiency and leads to unnecessary overhead due to GL calls transmission from Android space to Anbox runtime.
```

adglkh (changes requested):
@morphis I walked through this chapter and didn't find a section that covers the `swrast` platform, which is mesa + LLVMPipe backend for software only rendering. Probably we need to mention that on this page too?

exp/rendering-architecture.md
```diff
@@ -0,0 +1,54 @@
...
+
+For communication between the hardware composer module on the Android side and Anbox runtime, we use [Wayland](https://wayland.freedesktop.org/). So Anbox Cloud really functions as a compositor for Android i.e., the hardware composer module receives frames from the SurfaceFlinger and notifies Anbox runtime using Wayland. The Anbox runtime then submits the frame towards its output, which is either the screen or the streaming component.
+
+## For NVIDIA
+
+![Anbox Cloud NVIDIA pipeline|690x440](https://assets.ubuntu.com/v1/3ba1fddd-NVIDIA_pipeline.png)
+
+This NVIDIA pipeline uses the EmuGL stack. EmuGL is the hardware GLES emulation that is implemented with a mix of translator libraries, system libraries, and other rendering components. To understand more about EmuGL, see [EmuGL design overview](https://android.googlesource.com/platform/external/qemu/+/6654d90be28d0059baf854caccd4caf9f7033ccb/android/android-emugl/DESIGN).
```
morphis, line 38: We should not mention "EmuGL" or reference the Android docs but more describe the general concept of the API translation.

exp/rendering-architecture.md
```diff
@@ -0,0 +1,54 @@
...
+
+This NVIDIA pipeline uses the EmuGL stack. EmuGL is the hardware GLES emulation that is implemented with a mix of translator libraries, system libraries, and other rendering components. To understand more about EmuGL, see [EmuGL design overview](https://android.googlesource.com/platform/external/qemu/+/6654d90be28d0059baf854caccd4caf9f7033ccb/android/android-emugl/DESIGN). 
+
+For NVIDIA, we cannot use the NVIDIA driver inside the Android container because of compatibility issues. Hence, we use the Enterprise Ready NVIDIA driver that is available on every Ubuntu installation. So by using the EmuGL stack, we have an Anbox Cloud GPU driver which is a standard OpenGL ES or EGL driver that receives the API calls and converts them to remote procedure calls to the NVIDIA driver. The actual rendering and actions on the NVIDIA driver happens on the Anbox runtime side inside the Ubuntu container and not in the Android space.
+
+In terms of performance, compared to how rendering is done on Intel or AMD GPUs using Vulkan on top of ANGLE, using the EmuGL stack is suboptimal as it leads to unnecessary overhead due to transmission of GL calls from the Android space to Anbox runtime.
```
morphis, line 42: Can we word this a bit lighter in the sense of "... leads to very minimal additional overhead". Right now it sounds like its making things performing really bad which it by no means does. It's performant and and serves the need of cloud gaming with no compromise. Only certain operations are going to slower than with the direct approach (e.g. `glReadPixels`, but this one is slow anyway).

morphis (changes requested):

morphis:
> @morphis I walked through this chapter and didn't find a section that covers the `swrast` platform, which is mesa + LLVMPipe backend for software only rendering. Probably we need to mention that on this page too?

Good point but we can put that in another iteration as a follow up story (same as `null`).

keirthana:
> > @morphis I walked through this chapter and didn't find a section that covers the `swrast` platform, which is mesa + LLVMPipe backend for software only rendering. Probably we need to mention that on this page too?
> 
> Good point but we can put that in another iteration as a follow up story (same as `null`).

I have created a JIRA story for the same.

exp/rendering-architecture.md
keirthana: Okay, we are already explaining in simple terms what exactly EmuGL does - take in API calls and do remote procedure calls to the driver. So I have removed the terminology and the link.

adglkh:
LGTM

ru-fu:
> The inclusive naming check fails because of a URL that contains 'master' in its path. I will have to ignore it because it is an external URL that we are just referring to and we don't have any control on it.

Tip if you need it again some other time: You can ignore such errors by adding a `<!-- wokeignore:rule=master -->` comment on the same line as the violation.

exp/rendering-architecture.md
```diff
@@ -0,0 +1,52 @@
...
+# Supported GPU drivers
+
+For NVIDIA GPUs, Anbox Cloud uses the [Enterprise Ready Driver (ERD) from NVIDIA](https://help.ubuntu.com/community/NvidiaDriversInstallation) for Linux as available in Ubuntu. 
+For AMD and Intel GPUs, Anbox Cloud uses the [Mesa driver](https://www.mesa3d.org/).
+
+See [Component versions](https://anbox-cloud.io/docs/component-versions) to refer to the actual version supported for any particular Anbox Cloud release.
+
+# Supported APIs
+
+For NVIDIA GPUs Anbox Cloud uses OpenGL ES 3.2 and EGL 1.5 graphical interfaces. For Intel and AMD GPUs, Anbox Cloud uses Vulkan 1.3.
```
morphis, line 20: Should we link Vulkan to https://vulkan.org/ and OpenGL / EGL to https://www.khronos.org/opengles/ / https://www.khronos.org/egl/ ?

exp/rendering-architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+# Rendering architecture
+
+For NVIDIA GPUs, Anbox Cloud driver stacks currently have a native OpenGL driver. For AMD and Intel GPUs, an OpenGL or EGL driver is layered on top of the Vulkan API. Since Vulkan API provides better GPU management at a lower level than OpenGL or EGL, this approach is beneficial and preferred by many users.
```
morphis, line 24:
```suggestion
For NVIDIA GPUs, Anbox Cloud driver stacks currently have a native OpenGL ES driver. For AMD and Intel GPUs, an OpenGL ES or EGL driver is layered on top of the Vulkan API. Since Vulkan API provides better GPU management at a lower level than OpenGL ES or EGL, this approach is beneficial and preferred by many users.
```

exp/rendering-architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+## For NVIDIA
+
+![Anbox Cloud NVIDIA pipeline|690x440](https://assets.ubuntu.com/v1/3ba1fddd-NVIDIA_pipeline.png)
+
+For NVIDIA, as we cannot use the NVIDIA driver inside the Android container because of compatibility issues, we use the Enterprise Ready NVIDIA driver that is available on every Ubuntu installation. Instead, we have an Anbox Cloud GPU driver which is a standard OpenGL ES or EGL driver that receives the API calls and converts them to remote procedure calls to the NVIDIA driver. The actual rendering and actions on the NVIDIA driver happens on the Anbox runtime side inside the Ubuntu container and not in the Android space.
+
+In terms of performance, this could be perceived to have some transmission overhead when compared to the rendering on Intel and AMD GPUs. However, Anbox Cloud is optimised to keep this overhead minimal and the additional overhead due to the transmission of OpenGL calls from the Android space to Anbox runtime is not significant enough to affect most use cases.
```
morphis, line 40:
```suggestion
In terms of performance, this could be perceived to have some transmission overhead when compared to the rendering on Intel and AMD GPUs. However, Anbox Cloud is optimised to keep this overhead minimal and the additional overhead due to the transmission of OpenGL ES calls from the Android space to Anbox runtime is not significant enough to affect most use cases.
```

exp/rendering-architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+## For Intel and AMD
+
+![Anbox Cloud Intel and AMD pipeline|690x440](https://assets.ubuntu.com/v1/7aa9aff1-Intel_AMD_pipeline.png)
+
+For AMD and Intel GPUs, Anbox Cloud uses Vulkan as API in the Android space and we use ANGLE on top of Vulkan to circumvent OpenGL and EGL. Since the Mesa driver (vendor GPU driver) is available directly in the Android space, we do not have the overhead of the remote procedure call implementation as in the pipeline for NVIDIA.
```
morphis, line 46:
```suggestion
For AMD and Intel GPUs, Anbox Cloud uses Vulkan as API in the Android space and we use ANGLE on top of Vulkan to circumvent OpenGL ES and EGL. Since the Mesa driver (vendor GPU driver) is available directly in the Android space, we do not have the overhead of the remote procedure call implementation as in the pipeline for NVIDIA. 
```

morphis (changes requested):

PR #232, opened 2023-05-02
Fix incorrect escaping

keirthana:
Fix incorrect escaping of | character
JIRA link : https://warthogs.atlassian.net/browse/AC-1571


ajanon (approved):
That looks good to me as-is, let's just hope the discourse Markdown implementation reacts in the same way as the GitHub one does.

morphis (approved):
LGTM, thanks!

PR #240, opened 2023-05-15
Add information about scaling down LXD cluster

keirthana:
Although the [bug](https://bugs.launchpad.net/anbox-cloud/+bug/2019318) is about including the unschedulable configuration, the big picture of manual LXD upgrade should involve the full procedure of scaling down a LXD cluster.

So instead of just repeating the information about unschedulable configuration in the upgrade process and then redirecting to the "How to scale down a LXD cluster" topic, I have added a pointer to the topic if the user wishes to scale down.


ajanon (approved):
LGTM

morphis (approved):
LGTM

PR #247, opened 2023-05-31
Drop the one caveat which is no longer invalid

adglkh:
The issue to the deletion of LXD database node has been fixed in
dqlite/LXD and the LXD database node deletion has been supported
for quite a while.
Running the lxd-autoscaler for a while looks good.


keirthana (approved):
LGTM!

morphis (approved):
LGTM

PR #251, opened 2023-06-02
Clarify application versions

keirthana:
When updating an application, clarify wording around application versions being available and accessible to users for launch and use.

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -124,7 +124,6 @@ iptables
 init
 interprocess
 Javascript
-JIRA
 jitter
 JS
 JSON
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -124,6 +124,7 @@ iptables
 init
 interprocess
 Javascript
+JIRA
 jitter
 JS
 JSON
--- a/howto/application/update.md
+++ b/howto/application/update.md
@@ -84,7 +84,7 @@ In addition, each version has a status which indicates the status of the bootstr
 <a name="publish-application-versions"></a>
 ## Publish application versions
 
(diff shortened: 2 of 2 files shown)


.wordlist.txt
```diff
@@ -124,6 +124,7 @@ iptables
 init
 interprocess
 Javascript
+JIRA
```
ajanon, line 127: What is this for?

howto/application/update.md
```diff
@@ -84,7 +84,7 @@ In addition, each version has a status which indicates the status of the bootstr
 <a name="publish-application-versions"></a>
 ## Publish application versions
 
-The most important part of an application version is the `published` field. If a version is marked as published, it is accessible to users of Anbox Cloud. Generally when launching containers by using the AMS REST API, if no specific application version is given, the last published version of an application is used to create the container.
+The most important part of an application version is the `published` field. If a version is marked as published, it is available to launch and use. Generally when launching containers by using the AMS REST API, if no specific application version is given, by default, the last published version of an application is used to create the container.
```
ajanon, line 87: Shouldn't it be 'latest'?
```suggestion
The most important part of an application version is the `published` field. If a version is marked as published, it is available to launch and use. Generally when launching containers by using the AMS REST API, if no specific application version is given, by default, the latest published version of an application is used to create the container.
```

What do you think?

adglkh (approved):
Regarding the term `last` vs `latest`, not sure which one we all agree on in the end. I am fine with either.

.wordlist.txt
keirthana: Oops, I think this was a change meant for the other release notes PR and it crept it here. I will revert this in this PR as it is out of context.

howto/application/update.md
keirthana: Both are fine but in the context of versions, I would agree that 'latest' can be favoured.

PR #252, opened 2023-06-05
List the bug-fixing and improvement has been done for 1.18.1 release

adglkh:


announcements/1.18.1.md
```diff
@@ -4,16 +4,24 @@ The Anbox Cloud team is pleased to announce the release of Anbox Cloud 1.18.1.
...
 
 < List known issues with the current release. Include plans to fix it, if known. >
 
 ## Bug fixes
 
-< List bugs that were part of the current release. Include JIRA or launchpad IDs and links for the bugs >
+* [AC-1558](https://warthogs.atlassian.net/browse/AC-1558) Fix the crash caused by the `boot-application-relaunced` watchdog when launching Beach Buggy Racing in the portrait mode.
+* [AC-1600](https://warthogs.atlassian.net/browse/AC-1600) Fix the permission denial error when broadcasting `ACTION_CLOSE_SYSTEM_DIALOGS` intent from calendar app.
+* [AC-1606](https://warthogs.atlassian.net/browse/AC-1606) Fix the image ID out-of-sync issue when adding an image into AMS.
+* [AC-1608](https://warthogs.atlassian.net/browse/AC-1608) Fix the `InputReader: Received unexpected event` error when enabling the touch emulation by default.
```
ajanon, line 21: Isn't that the same as above?
> Do not send the `touch-move` event until the `touch-start` event is fired first. This avoid unexpected events sent from the client to Android container, which causes a bunch of error logs to show up during Android runtime.

Do we need to have it in both sections? What do you think?

adglkh: Let me drop it from the `## New features & improvements` seciton then

PR #256, opened 2023-06-09
Restructure release notes page

keirthana:
* Move content from release notes page to `release-version.md` files for all versions
* Add navigation table with no level entries for all versions from current version through version 1.7.1. These are the versions that have corresponding discourse posts
* Rewrite release notes page to contain links to release notes till version 1.7.1 and then archive release notes for versions 1.7.0 through to 1.0.0 under an expandable content block

**Note**: The spellchecker will fail as a lot of old files have words that are not added to the ignore list. It is not the best use of time to go back and fix all those. 

Screenshots attached to this pull request show how the final release notes page will look.

### DIFF
--- a/index.md
+++ b/index.md
@@ -176,6 +176,55 @@ Thinking about using Anbox Cloud for your next project? [Get in touch!](https://
 | 2 | roadmap | [Release roadmap](https://discourse.ubuntu.com/t/release-roadmap/19359)|
 | 2 | supported-versions | [Supported versions](https://discourse.ubuntu.com/t/supported-versions/21046)|
 | 2 | component-versions | [Component versions](https://discourse.ubuntu.com/t/component-versions/21413)|
+|   | release-notes/1.18.0 | [Release notes-Anbox Cloud 1.18.0](https://discourse.ubuntu.com/t/35812)|
+|   | release-notes/1.17.2 | [Release notes-Anbox Cloud 1.17.2](https://discourse.ubuntu.com/t/35195)|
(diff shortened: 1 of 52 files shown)


keirthana:
**Note**: The spellchecker will fail as a lot of old files have words that are not added to the ignore list. It is not the best use of time to go back and fix all those. 

The following screenshots show snaps of how the final release notes page will look.
**Release notes page at first look:**
first_look_rn_page

**Everything except recent releases is combined into an expandable content block:**
expand-older-releases

**Topics built from discourse posts till 1.7.1:**
till-1 7 1

**Content retained under expandable content block prior to 1.7.1:**
expand-prior-to-1 7 1

adglkh:
I like this new way for people to browser all release versions.
1. Do we have a year `2021` under the Old release expandable content block?
2. It may be my personal preference. It'd be great if there is an `Upcoming release` that only lists a single release version with no URL refer to above the `Recent releases`, so people could know what's the next release version without jumping to Release roadmap

keirthana:
> I like this new way for people to browser all release versions.
> 
>     1. Do we have a year `2021` under the Old release expandable content block?
> 
>     2. It may be my personal preference. It'd be great if there is an `Upcoming release` that only lists a single release version with no URL refer to above the `Recent releases`, so people could know what's the next release version without jumping to Release roadmap

1. Yes, we do. it is in the `release-notes.md` file.
2. This is good idea. Let me add that.

adglkh (approved):
LGTM (one nitpick though)

PR #260, opened 2023-06-20
Revert unschedulable for PATCH method of 1.0/nodes endpoint

keirthana:
Based on a bug report, we found out that the `unschedulable` correction was not done for `type NodePatch struct`. A [Jira ticket](https://warthogs.atlassian.net/browse/AC-1689) is tracking this issue. For now, the docs need to reflect the field name correctly.

### DIFF
--- a/ref/ams-http-api.md
+++ b/ref/ams-http-api.md
@@ -2044,7 +2044,7 @@ Supported update field            |   Description
 `gpu_slots` | (number) Slots on the GPU available to containers
 `gpu_encoder_slots` | (number) Slots on the GPU encoders available to containers
 `tags` | (string) User defined tags
-`unschedulable` | (Boolean) Whether this LXD node is schedulable by AMS
+`unscheduable` | (Boolean) Whether this LXD node is schedulable by AMS. (The typo in the field name will be fixed in an upcoming release.)
 
 A sample payload as follows:
 


ref/ams-http-api.md
```diff
@@ -2044,7 +2044,7 @@ Supported update field            |   Description
 `gpu_slots` | (number) Slots on the GPU available to containers
 `gpu_encoder_slots` | (number) Slots on the GPU encoders available to containers
 `tags` | (string) User defined tags
-`unschedulable` | (Boolean) Whether this LXD node is schedulable by AMS
+`unscheduable` | (Boolean) Whether this LXD node is schedulable by AMS. (The typo in the field name will be fixed in an upcoming release.)
```
adglkh, line 2047: Let's drop the `(The typo in the field name will be fixed in an upcoming release.)` since for the backport compatibility concern, even if we fix it in an upcoming release, `unschedulable` would still be supported next to `unscheduable`, we're not gonna drop it.

adglkh (changes requested):

ref/ams-http-api.md
keirthana: @adglkh Originally we had this note, not in every instance though. We had it on the AMS configuration page. This is just to stop the user from wondering that whether this is a typo just in this instance because everywhere else it is `unschedulable` on the docs. If you still want to drop the note, let me know.

adglkh: Okay. let's keep it then.

adglkh (approved):
LGTM

PR #261, opened 2023-06-21
Fix broken links

keirthana:
* Fix broken links as automatically reported on ~docs mattermost channel

We can currently ignore discourse links vs docs links usage in release notes files. There is a mix of both. While discourse links help in reducing broken links, docs links redirect to the correct output format when users are reading the release notes. We need to come up with a strategy for this but that's a task for later and also best done iteratively as we edit files.

### DIFF
--- a/howto/application/test.md
+++ b/howto/application/test.md
@@ -82,7 +82,8 @@ If the APK file that you want to test is located in local folder or hosted on a 
 The above preset accepts either an absolute path to an APK or a URL.
 
 Finally you can start a new test session and run your test cases.
-For more details about Appium, please refer to the [official documentation](http://appium.io/docs/en/about-appium/getting-started/)
+
+For more details about Appium, please refer to the [Appium official documentation](http://appium.io)
 
 ### APK managed by AMS
 
--- a/howto/application/test.md
+++ b/howto/application/test.md
@@ -83,7 +83,7 @@ The above preset accepts either an absolute path to an APK or a URL.
 
 Finally you can start a new test session and run your test cases.
 
-For more details about Appium, please refer to the [Appium official documentation](http://appium.io)
+For more details about Appium, please refer to the [Appium official documentation](https://appium.io)
 
 ### APK managed by AMS
 
--- a/howto/container/access.md
+++ b/howto/container/access.md
@@ -30,8 +30,8 @@ In the next sections, you will learn how to connect a container running on Anbox
 ### Install scrcpy
 
 Scrcpy is not available from the official Ubuntu repositories. Therefore, you must install scrcpy in one of the following ways:
-
-* Build scrcpy from the source by following the official [guide](https://github.com/Genymobile/scrcpy/blob/master/BUILD.md)
+<!-- wokeignore:rule=master -->
+* Build scrcpy from the source by following the official [guide](https://github.com/Genymobile/scrcpy/blob/master/doc/build.md)
 
 * Install scrcpy from the [Ubuntu snap store](https://snapcraft.io):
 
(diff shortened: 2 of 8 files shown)


howto/application/test.md
```diff
@@ -82,7 +82,8 @@ If the APK file that you want to test is located in local folder or hosted on a
 The above preset accepts either an absolute path to an APK or a URL.
 
 Finally you can start a new test session and run your test cases.
-For more details about Appium, please refer to the [official documentation](http://appium.io/docs/en/about-appium/getting-started/)
+
+For more details about Appium, please refer to the [Appium official documentation](http://appium.io)
```
ajanon, line 86:
```suggestion
For more details about Appium, please refer to the [Appium official documentation](https://appium.io)
```

It doesn't change much, but I think it's better to default to a secure connection.

howto/container/access.md
```diff
@@ -30,8 +30,8 @@ In the next sections, you will learn how to connect a container running on Anbox
 ### Install scrcpy
 
 Scrcpy is not available from the official Ubuntu repositories. Therefore, you must install scrcpy in one of the following ways:
-
-* Build scrcpy from the source by following the official [guide](https://github.com/Genymobile/scrcpy/blob/master/BUILD.md)
+<!-- wokeignore:rule=master -->
+* Build scrcpy from the source by following the official [guide](https://github.com/Genymobile/scrcpy/blob/master/doc/build.md)
```
ajanon, line 34: Isn't it better to have a permalink instead, to avoid similar issues in the future?
```suggestion
* Build scrcpy from the source by following the official [guide](https://github.com/Genymobile/scrcpy/blob/4ad74794255b7669a89c023afc7ae7a2c0da4dc3/doc/build.md)
```

But then, this link may get outdated. It may be better to have a broken link as it may be easier to spot. What do you think?

ref/webrtc-streamer.md
```diff
@@ -17,7 +17,7 @@ The configuration file uses the JSON format. It has the following structure:
 | `nvidia_h264.aq_strength` | integer | `0` | Strength of adaptive quantisation: a value from `1` (least aggressive) to `15` (most aggressive). `0` means the encoder will automatically decide. |
 | `nvidia_h264.preset` | integer | `0` | Preset to use (a value from `1` to `7`). `0` means Anbox will automatically decide. |
 
-See the [NVENC Video Encoder API](https://docs.nvidia.com/video-technologies/video-codec-sdk/nvenc-video-encoder-api-prog-guide/) documentation for more details on the `nvidia_h264` options.
+See the [NVENC Video Encoder API](https://docs.nvidia.com/video-technologies/video-codec-sdk/12.1/nvenc-video-encoder-api-prog-guide/index.html) documentation for more details on the `nvidia_h264` options.
```
ajanon, line 20: The `index.html` part seems unnecessary to me: it works without it. Also, I think we are using the 11.0.10 version of NVEnc.

@adglkh Am I correct? The build files include:
```
external/nv-video-codec-sdk/11.0.10
```

```suggestion
See the [NVENC Video Encoder API](https://docs.nvidia.com/video-technologies/video-codec-sdk/11.0/nvenc-video-encoder-api-prog-guide/) documentation for more details on the `nvidia_h264` options.
```

ajanon (changes requested):

howto/container/access.md
keirthana: That's exactly what I was thinking, we can easily spot the broken link with the link checker. I think we can retain it for now and see how often it changes. If we end up fixing a broken link too often, then the permalink might be worth using.

ajanon (approved):
LGTM

PR #263, opened 2023-06-23
Add a pre-commit hook for checking doc links

keirthana:
Add a pre-commit hook that finds links pointing to `https://anbox-cloud.io/docs/...` instead of discourse and reports errors.


ajanon (approved):
LGTM

PR #277, opened 2023-07-21
Create a tutorial for setting up a streaming client

keirthana:
Pending decisions/changes:
* Do we do away with the how to guide or refine them to be less detailed?
* Link this tutorial to other topics where relevant
* Depending on the answer to first point, update links where the how to guide is referred to point to the tutorial link

### DIFF
--- a/tut/stream-client.md
+++ b/tut/stream-client.md
@@ -17,7 +17,7 @@ The output of this command provides the access token. Make a note of this token 
 
 ### Create an application
 
-Follow the instructions in [How to create an application](https://discourse.ubuntu.com/t/24198) and create an application.
+Follow the instructions in [How to create an application](https://discourse.ubuntu.com/t/24198) and create an application. You can create any application, the application type does not matter.
 
 ## Set up the stream client
 Create a directory to set up the stream client:
@@ -42,14 +42,14 @@ Create a `demo.html` file inside `/srv/stream-client` with the following content
         // Run `juju status` to get the stream gateway IP/domain name
         // Replace 'https://gateway.url.net' with the stream gateway IP/domain name
         url: 'https://gateway.url.net',
-        //Use the access token created earlier as 'YOUR_AUTH_TOKEN'
+        // Use the access token created earlier as 'YOUR_AUTH_TOKEN'
         authToken: 'YOUR_AUTH_TOKEN',
         session: {
             // Run `amc list` to see a list of all applications
             // Replace `com.foo.bar` with your application name
             app: "com.foo.bar",
         },
-        // Set the display resolution and frame rate for the Android container
+        // Adjust the display resolution and frame rate for the Android container as per your requirements
         screen: {
             width: 1280,
             height: 720,
@@ -76,11 +76,13 @@ Create a `demo.html` file inside `/srv/stream-client` with the following content
     <div id="anbox-stream" style="width: 100vw; height: 100vh;"></div>
 </body>
 ```
-If you experience any streaming issues, you can turn on debug information by adding the following option to the `AnboxStream` which is the `main` class:
+If you experience any streaming issues, you can turn on debug information by adding the following option to the `AnboxStream`:
 
     experimental: {
        debug: true,
-       }
+    }
+
+See [how to troubleshoot streaming issues](https://discourse.ubuntu.com/t/31341) for common issues related to streaming.
 
 ## Create and enable the stream UI service
 
@@ -99,17 +101,26 @@ WantedBy=multi-user.target
 
 Run the following commands to install and start the service:
 ```bash
-#Install the service file
+# Install the service file
 sudo cp stream-ui.service /etc/systemd/system/
-#Reload systemd
+# Reload systemd
 sudo systemctl daemon-reload
-#Enable and start the service
+# Enable and start the service
 sudo systemctl enable --now stream-ui.service
 ```
+## Generate the HTTP basic authentication credentials
+
+Use the `htpasswd` tool to generate a user/password combination:
+```
+apt install -y apache2-utils
+htpasswd -n <your user name>
+```
+
+Enter your desired password when prompted.
 
 ## Add a middleware definition
 
-Add a middleware definition (see [Adding Basic Authentication](https://doc.traefik.io/traefik/v2.0/middlewares/basicauth/) for more details) to the [`traefik`](https://traefik.io/) configuration.
+The appliance uses [`traefik`](https://traefik.io/) as the reverse proxy for routing incoming requests and you need to add a middleware definition for the stream client to the `traefik` configuration. See [Adding Basic Authentication](https://doc.traefik.io/traefik/v2.0/middlewares/basicauth/) for more details.
 
 Create `stream-ui.yaml` under `/var/snap/anbox-cloud-appliance/common/traefik/conf/` with the following content for `traefik` to redirect requests to the service. 
 
@@ -126,8 +137,9 @@ http:
     demo-auth:
       basicAuth:
         users:
-        # demo is the user name and
-        # foobar is the password for HTTP basic authentication
+        # Replace 'demo' with the user name and
+        # 'foobar' with the password generated for HTTP basic 
+        # authentication earlier with the 'htpasswd' tool
         - "demo:foobar"
   services:
     stream-ui-demo:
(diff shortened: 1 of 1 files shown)


tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
+This tutorial guides you through the process of setting up a web-based streaming client using the Anbox Cloud streaming stack.
+
+## Preparation
```
ajanon, line 3: Setup? Prerequisites?
I'm not against Preparation, I just wonder if something else would sound better.

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+Run the following commands to install and start the service:
+```bash
+#Install the service file
+sudo cp stream-ui.service /etc/systemd/system/
+#Reload systemd
+sudo systemctl daemon-reload
+#Enable and start the service
+sudo systemctl enable --now stream-ui.service
```
ajanon, line 107:
```suggestion
# Install the service file
sudo cp stream-ui.service /etc/systemd/system/
# Reload systemd
sudo systemctl daemon-reload
# Enable and start the service
sudo systemctl enable --now stream-ui.service
```

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+    stream-ui-demo:
+      loadBalancer:
+        servers:
+          - url: http://127.0.0.1:8080
+```
+
+Set the right permissions for the `stream-ui.yaml`:
+
+    chmod 0600  /var/snap/anbox-cloud-appliance/common/traefik/conf/stream-ui.yaml
+
```
ajanon, line 142:
````suggestion
```bash
chmod 0600 /var/snap/anbox-cloud-appliance/common/traefik/conf/stream-ui.yaml
```
````

Or keep it as-is, but please remove the superfluous space:
```suggestion

    chmod 0600 /var/snap/anbox-cloud-appliance/common/traefik/conf/stream-ui.yaml

```

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+## Generate the HTTP basic authentication credentials
+
+Use the `htpasswd` tool to generate a user/password combination:
+```
+apt install -y apache2-utils
+httpasswd -n <your user name>
```
ajanon, line 148:
```suggestion
htpasswd -n <your user name>
```

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+    demo-auth:
+        basicAuth:
+            users:
+            - "<user name>:<hashed password>"
+...
+```
+
+With this configured, users will be asked to enter the basic authentication credentials to access the site.
+
+Restart `traefik`:
```
ajanon, line 170:
```suggestion
Restart `traefik` for the configuration changes to take effect:
```
What do you think?

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+    });
+    stream.connect();
+    </script>
+    <!--To display the video element correctly, always specify both the height and the width attributes-->
+    <div id="anbox-stream" style="width: 100vw; height: 100vh;"></div>
+</body>
+```
+If you experience any streaming issues, you can turn on debug information by adding the following option to the `AnboxStream` which is the `main` class:
```
ajanon, line 79:
```suggestion
If you experience any streaming issues, you can turn on debug information by adding the following option to the `AnboxStream`:
```

The comment about the `main` class does not seem necessary to me.

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+
+    experimental: {
+       debug: true,
+       }
```
ajanon, line 83:
```suggestion
    experimental: {
       debug: true,
    }
```

ajanon (changes requested):

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+
+On the machine where Anbox Cloud Appliance is installed, create the service account by running the following command:
+
+    anbox-cloud-appliance gateway account create test
+
+The output of this command provides the access token. Make a note of this token to use when you make a request to the stream gateway API.
+
+### Create an application
+
+Follow the instructions in [How to create an application](https://discourse.ubuntu.com/t/24198) and create an application.
```
morphis, line 20: Should we leave a note that the type of application (game, ..) doesn't matter?

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+<body>
+    <script type="module">
+    import {AnboxStream, AnboxStreamGatewayConnector} from './anbox-stream-sdk.js';
+
+    const connector = new AnboxStreamGatewayConnector({
+        // Run `juju status` to get the stream gateway IP/domain name
+        // Replace 'https://gateway.url.net' with the stream gateway IP/domain name
+        url: 'https://gateway.url.net',
+        //Use the access token created earlier as 'YOUR_AUTH_TOKEN'
+        authToken: 'YOUR_AUTH_TOKEN',
```
morphis, line 46: Should we point out that URL, token and app name need to be replaced?

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+        session: {
+            // Run `amc list` to see a list of all applications
+            // Replace `com.foo.bar` with your application name
+            app: "com.foo.bar",
+        },
+        // Set the display resolution and frame rate for the Android container
+        screen: {
+            width: 1280,
+            height: 720,
+            fps: 25,
```
morphis, line 56: Should we also indicate that these can be adjusted?

tut/stream-client.md
morphis: Shouldn't we also point to our troubleshooting guides?

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+sudo cp stream-ui.service /etc/systemd/system/
+#Reload systemd
+sudo systemctl daemon-reload
+#Enable and start the service
+sudo systemctl enable --now stream-ui.service
+```
+
+## Add a middleware definition
+
+Add a middleware definition (see [Adding Basic Authentication](https://doc.traefik.io/traefik/v2.0/middlewares/basicauth/) for more details) to the [`traefik`](https://traefik.io/) configuration.
```
morphis, line 112: Should we explain somewhere that the appliance uses traefik as component and we can reuse it for the purpose of the tutorial here?

tut/stream-client.md
```diff
@@ -0,0 +1,174 @@
...
+      service: stream-ui-demo
+      priority: 110
+      tls: {}deployed
+        forceSlash: false
+    demo-auth:
+      basicAuth:
+        users:
+        # demo is the user name and
+        # foobar is the password for HTTP basic authentication
+        - "demo:foobar"
```
morphis, line 131: This is not correct. The password need to be generated we `htpasswd ..`

morphis (changes requested):

tut/stream-client.md
keirthana: I have indicated it using the comments hoping it would be a step by step guide. Do you think we need to reiterate using explanation text?

tut/stream-client.md
keirthana: I have done so now in comments, we can take the same approach as for the earlier URL, token and app name in case we go with text explanation

tut/stream-client.md
keirthana: Added an introductory line for this.

tut/stream-client.md
keirthana: Rearranged the order of generating the password and adding it to the configuration a bit so that it does not sound so confusing

tut/stream-client.md
ajanon: Makes sense!

tut/stream-client.md
```diff
@@ -0,0 +1,159 @@
...
+
+    experimental: {
+       debug: true,
+    }
+
+See [how to troubleshoot streaming issues](https://discourse.ubuntu.com/t/31341) for common issues related to streaming.
+
+## Create and enable the stream UI service
+
+Create a `stream-ui.service` with the following content:
```
ajanon, line 89:
```suggestion
Create a `stream-ui.service` file with the following content:
```
Or
```suggestion
Create `stream-ui.service` with the following content:
```

tut/stream-client.md
```diff
@@ -0,0 +1,159 @@
...
+```
```
ajanon, line 90:
````suggestion
```service
````

ajanon (changes requested):
Two nitpicks.

tut/stream-client.md
keirthana: Good catch, thanks!

tut/stream-client.md
keirthana: I am not sure if the discourse to docs formatting shows any different formatting for different types of code. However, the preview in VS code does for most types (not for service though). I think this is a best practice in any case even if there is no difference in the output. Thanks!

ajanon (approved):
LGTM

tut/stream-client.md
morphis: I think we should give a hint to the user that he needs to replace something. Otherwise the sentence `Create a `demo.html` file inside `/srv/stream-client` with the following content:` can be taken as that and the user simply copies the file and does not read through it.

morphis (changes requested):

tut/stream-client.md
keirthana: Have added a note along with removing the "following content" part to avoid the copy-paste temptation.

morphis: That's perfect, thanks!

morphis (approved):

PR #280, opened 2023-07-28
Remove stream client howto topic

keirthana:
Move the content regarding SDKs in the howto guide of stream client and remove the topic.
Retaining the "How to access stream gateway" topic as it has a lot of additional information.


tut/stream-client.md
```diff
@@ -74,7 +79,8 @@ Create a `demo.html` file inside `/srv/stream-client`:
     });
     stream.connect();
     </script>
-    <!--To display the video element correctly, always specify both the height and the width attributes-->
+    <!--To display the video element correctly,
+    always specify both the height and the width attributes-->
```
ajanon, line 83:
```suggestion
    <!--To display the video element correctly,
        always specify both the height and the width attributes-->
```
To make the text aligned. It's not necessary, I just think it looks a bit better, but I'm fine either way. What do you think?

ajanon (changes requested):
One nitpick

morphis (approved):
LGTM, thanks!

howto/stream/access.md
```diff
@@ -22,17 +19,9 @@ If you are running the Anbox Cloud Appliance, use the following command:
 
     anbox-cloud-appliance gateway account create my-client
 
-**Deleting a token**
+### Using the token in your requests
```
adglkh, line 22: Maybe simplified to `Using the token` to correlate with `Creating a token` in the above and `Deleting a token` below. To me, if we want to keep it as is, `Creating a token for your request` sounds more natural.

adglkh (approved):
LGTM

PR #282, opened 2023-08-02
Move AV1 encoding from future to supported

keirthana:
This change should be published only with 1.19 updates.

### DIFF
--- a/ref/supported-video-codecs.md
+++ b/ref/supported-video-codecs.md
@@ -6,10 +6,8 @@ Currently, the following video codecs are supported:
 
  * H.264 - The use of H.264 requires a license from the [MPEG LA](https://www.mpegla.com/). Ensure you have the rights to stream H.264 encoded video content to your users.
  * VP8
-
-In the future we plan to add support for:
-
- * VP9
  * AV1
 
+In the future we plan to add support for VP9 as well.
+ 
 Availability of additional codecs depends on them being supported by the GPU vendors in their hardware encoding solutions or if a viable software encoding solution exists. See [Release roadmap](https://discourse.ubuntu.com/t/19359) for future versions of Anbox Cloud and planned features.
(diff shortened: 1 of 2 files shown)


ajanon (approved):
LGTM

ref/supported-video-codecs.md
```diff
@@ -6,10 +6,8 @@ Currently, the following video codecs are supported:
 
  * H.264 - The use of H.264 requires a license from the [MPEG LA](https://www.mpegla.com/). Ensure you have the rights to stream H.264 encoded video content to your users.
  * VP8
-
-In the future we plan to add support for:
-
- * VP9
  * AV1
 
+In the future we plan to add support for VP9 as well.
```
adglkh, line 11: I am not sure if we want to explicitly call out here in the doc. But the following looks better:
```
* AV1 - Depending on the GPU model used in the deployment, AV1 hardware encoding is the default preference over H.264 on supported GPUs, such as NVIDIA Ada Lovelace Architecture-based GPUs like L4. Otherwise, it will fallback to AV1 software encoding.
```

adglkh (commented):
In the ref/sdks.md file, Under the `Features` table, we need to add `AV1` codec as well for `Supported video codecs`

ref/supported-video-codecs.md
keirthana: Good idea, I have added it along with the supported video codes list information.

keirthana:
> In the ref/sdks.md file, Under the `Features` table, we need to add `AV1` codec as well for `Supported video codecs`

Added, good catch!

adglkh (approved):
LGTM

ref/supported-video-codecs.md
```diff
@@ -5,11 +5,9 @@ Not all codecs are supported by one or more of the supported GPU models, neither
...
  * H.264 - The use of H.264 requires a license from the [MPEG LA](https://www.mpegla.com/). Ensure you have the rights to stream H.264 encoded video content to your users.
+ * AV1 - Depending on the GPU model used in the deployment, AV1 hardware encoding is the default preference over H.264 on supported GPUs, such as NVIDIA Ada Lovelace Architecture-based GPUs like L4. Otherwise, it will fallback to AV1 software encoding.
  * VP8
 
-In the future we plan to add support for:
-
- * VP9
- * AV1
-
+In the future we plan to add support for VP9 as well.
```
morphis, line 11: Lets drop this one. We currently do not plan anymore to eventually support VP9. With AV1 being supported there should be no need.

morphis (changes requested):

morphis (approved):
LGTM

PR #283, opened 2023-08-03
Remove unsupported images

keirthana:
* Removed bionic images and jammy android 10 images from the supported list
* Changed mentions of bionic to a more recent example.

<s>Pending query: The mentions of bionic in https://anbox-cloud.io/docs/howto/monitor/monitor-status. Can they be just changed to jammy because it is part of the nagios configuration?</s>

### DIFF
--- a/ref/provided-images.md
+++ b/ref/provided-images.md
@@ -18,10 +18,10 @@ The following table lists supported images available on the official image serve
 ## Deprecated images
 The following table lists the images that are available on the official image server but are deprecated. They will become unsupported after two releases.
 
-| Name                        | Android Version | Ubuntu Version | Available since |
-|-----------------------------|-----------------|----------------|---------------|
-| `jammy:android11:amd64`     | 11              | 22.04          | 1.14 |
-| `jammy:android11:arm64`     | 11              | 22.04          | 1.14 |
+| Name                        | Android Version | Ubuntu Version | Available since | Deprecated since |
+|-----------------------------|-----------------|----------------|-----------------|--------------------|
+| `jammy:android11:amd64`     | 11              | 22.04          | 1.14            | 1.19 |
+| `jammy:android11:arm64`     | 11              | 22.04          | 1.14            | 1.19 |
 
 ## Image support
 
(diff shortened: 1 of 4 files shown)


adglkh (approved):
LGTM. 
One question:
Will we call out in the doc that the Android 11 image is deprecated since 1.19 release and will be dropped in the 1.20 release?

keirthana:
> LGTM. One question: Will we call out in the doc that the Android 11 image is deprecated since 1.19 release and will be dropped in the 1.20 release?

I have mentioned the it generally without mentioning the release version numbers  under "Deprecated images".

ref/provided-images.md
```diff
@@ -1,36 +1,32 @@
...
+
+| Name                        | Android Version | Ubuntu Version | Available since |
+|-----------------------------|-----------------|----------------|---------------|
+| `jammy:android13:amd64`     | 13              | 22.04          | 1.16 |
+| `jammy:android13:arm64`     | 13              | 22.04          | 1.16 |
+| `jammy:android12:amd64`     | 12              | 22.04          | 1.14 |
+| `jammy:android12:arm64`     | 12              | 22.04          | 1.14 |
+
+## Deprecated images
+The following table lists the images that are available on the official image server but are deprecated. They will become unsupported after two releases.
```
ajanon, line 19: Should we also add _when_ they were deprecated? If somebody is reading this after 1.19 is released, how would they know if it was deprecated in 1.18 or 1.19?

ajanon (changes requested):

ref/provided-images.md
keirthana: Added a deprecated since column which would address this concern.

ajanon (approved):
LGTM

ref/provided-images.md
```diff
@@ -1,36 +1,32 @@
...
+
+| Name                        | Android Version | Ubuntu Version | Available since | Deprecated since |
+|-----------------------------|-----------------|----------------|-----------------|--------------------|
```
morphis, line 22: Should we add a column saying with which version (1.21) they get dropped?

morphis (changes requested):

ref/provided-images.md
keirthana: > Should we add a column saying with which version (1.21) they get dropped?

This is a commitment that we make in advance, so we can add it if we are absolutely sure that it will be dropped in 1.21.

morphis: That's fair

morphis (approved):
LGTM

PR #284, opened 2023-08-04
Remove Juju install mentions for appliance

keirthana:


ref/requirements.md
```diff
@@ -77,7 +77,7 @@ Anbox Cloud currently supports the following LXD versions:
 <a name="juju-version"></a>
 ### Juju version
 
-Anbox Cloud requires [Juju](https://juju.is/) to be installed to manage the different components and their dependencies. Starting with Anbox Cloud 1.13, the required Juju version is **2.9**. Earlier versions require Juju **2.8**.
+The regular Anbox Cloud variant requires [Juju](https://juju.is/) to be installed to manage the different components and their dependencies. Starting with Anbox Cloud 1.13, the required Juju version is **2.9**. Earlier versions require Juju **2.8**.
```
morphis, line 80:
```suggestion
The regular Anbox Cloud variant requires [Juju](https://juju.is/) to be installed to manage the different components and their dependencies. The minimum required Juju version is **2.9**.
```

Lets drop the mention of 2.8 entirely. We support/can work with 2.9+. CI actually tests 3.1 today by default.

morphis (changes requested):

tut/installing-appliance.md
```diff
@@ -66,12 +66,7 @@ Run the following command to install the `anbox-cloud-appliance` snap, which han
 <a name="additional-tools"></a>
 ### 5. Install additional tools
```
ajanon, line 67: Should we keep it in its own subsection or should we merge two subsections?

This may be better left to another PR if you want to keep it small in scope, though.

ajanon (approved):
LGTM, one comment

tut/installing-appliance.md
keirthana: No, I think it can stay in a separate section because it is a tutorial and it is useful for readers to understand that AMC is an additional tool we are installing on top of the appliance. Also, the additional tools sub section is referred in the getting started tutorial as a prereq.

morphis (approved):
LGTM

PR #293, opened 2023-08-23
Review reference topics

keirthana:


ref/addons.md
```diff
@@ -22,15 +21,16 @@ To create the addon, you must provide `amc` with either of the following:
...
-| `description`         | string       | Description of the addon.                                                                                                                                                                                                     |                                       |
-| `provides`            | object       | Additional capabilities that this addon provides to the container. See individual items for details.                                                                                                                          |                                       |
+| Name                  | Type         | Description    | Allowed values  |
+|-----------------------|--------------|----------------|-----------------|
+| `name`                | string       | Name of the addon. Can be used to reference the addon.| All characters except for the following: `< > : " / \ \| ? *`, as well as space. |
+| `description`         | string       | Description of the addon.   |        |
+| `provides`            | object       | Additional capabilities that this addon provides to the container. See individual items for details.          |             |
```
ajanon, line 28: Shouldn't you also add something in the 'Allowed values' column? I guess these two can contain any character (ASCII? UTF-8?), but we should maybe check.

ref/perf-benchmarks.md
```diff
@@ -1,21 +1,21 @@
...
+The benchmarks were performed using the `amc benchmark` utility as described in [How to run benchmarks](https://discourse.ubuntu.com/t/benchmarking-a-deployment/17770). The results describe the maximum number of parallel running containers (column **# Containers**) delivering a stable frame rate (column **Avg. FPS**). Running more containers either gives a very high variation in the provided frame rate or is not possible due to other hardware limitations such as system memory, GPU memory etc.
 
-All benchmarks include rendering and video encoding. On machines/VMs without a GPU, rendering and video encoding are performed in software on the CPU.
+For most of the benchmarks below, a special version of the [BombSquad](https://www.froemling.net/apps/bombsquad) application was used. This version runs in a demo mode in which the game provides random and automated simulated gameplay. This allowed the benchmark to simulate a real-world scenario where actual users would play the game, instead of deriving from a static game scene without much variation across the Android instances.
```
ajanon, line 5:
```suggestion
For most of the benchmarks below, a special version of the [BombSquad](https://www.froemling.net/apps/bombsquad) application was used. This version runs in a demo mode in which the game provides random and automated simulated gameplay. This allowed the benchmark to simulate a real-world scenario where actual users would play the game, instead of deriving from a static game scene without much variation.
```
I feel like 'the Android instances' at the end of the sentence is a bit confusing. What do you think?

ref/roadmap.md
```diff
@@ -8,8 +8,21 @@ Anbox Cloud follows a defined release cycle with frequent minor and patch releas
...
 |---------|-------------|---------------|
 | 1.19.1 | September 13 2023 | * Android security updates for September 2023<br/>* Bug fixes |
 | 1.19.2 | October 11 2023 | * Android security updates for October 2023<br/>* Bug fixes |
+| 1.20.0 | November 15 2023 |  |
+
+## Professional support
+
+If you are looking for additional support, find out about [Ubuntu Pro](https://ubuntu.com/support).
+
+Canonical can also provide [managed solutions](https://ubuntu.com/managed) for Anbox Cloud.
```
ajanon, line 28: Does that really fit in the roadmap document?

ajanon (changes requested):
Some comments

ref/roadmap.md
keirthana: I was moving it around to see if it is better to keep a separate topic for this or merge it. I didn't realise it got added to the commit. Thanks! I have reverted it and planning to keep it as is now.

ref/addons.md
keirthana: yes, I will check this and add relevant information.

ref/perf-benchmarks.md
keirthana: I dont' find it confusing but in any case, it does not add any value to the statement. I agree to removing it.

morphis (approved):
LGTM, good to see these improvements!

ajanon (approved):
LGTM!

ref/android-features.md
```diff
@@ -1,11 +1,11 @@
-Anbox Cloud implements support for various features that Android has. The following table provides an overview of which features are currently available as part of Anbox Cloud and which are not.
+Anbox Cloud implements support for various Android features. The following table lists certain Android features and if they are supported in Anbox Cloud.
 
 | Feature           | Supported | Notes  |
 |--------------------|-----------------|-------------|
 | [OpenGL ES](https://source.android.com/devices/graphics/arch-egl-opengl) <= 3.2   |  ✓  | |
 | [Vulkan](https://source.android.com/devices/graphics/arch-vulkan) |  ✓  | |
-| [Camera](https://source.android.com/devices/camera) |  ✓  | Currently not working for Intel and AMD GPUs. |
-| [Sensors](https://source.android.com/devices/sensors) |  ✓  | Currently, the sensors that a client device has are not connected to the Android container when streaming. |
+| [Camera](https://source.android.com/devices/camera) |  ✓  | Does not work with Intel and AMD GPUs. |
```
adglkh, line 7: Since 1.19 release, this is no longer invalid. And camera support should be working for the direct rendering cases where the Intel and AMD gpus are in used. @morphis could you confirm this?

howto/application/resources.md
```diff
@@ -1,23 +1,27 @@
...
 
-```
+```yaml
 name: my-application
 instance-type: a4.3
 resources:
-  disk-size: 5GB
+  cpus: 6
+  memory: 4GB
+  disk-size: 8GB
```
adglkh, line 13: The new example makes less sense to me in practice since if people customize each resource, the `instance-type` can be omitted here in the manifest file. If we really want to demonstrate how to override one or two certain resources that were given from `instance-type` with the `resources` attribute. I'd suggest dropping either `cpus` or `memory` in the above example.

ref/android-features.md
keirthana: Dropped the notes for Camera and Sensors as discussed in SU

ref/addons.md
keirthana: Dropped the `provides` as discussed in SU.

howto/application/resources.md
keirthana: Makes sense, modified as you suggested

PR #297, opened 2023-08-31
Fix spellcheck errors

keirthana:


adglkh (commented):
The empty line is needed here after `BEGIN PGP PUBLIC KEY BLOCK`, otherwise you'll see https://github.com/keybase/kbpgp/issues/198

keirthana:
> The empty line is needed here after `BEGIN PGP PUBLIC KEY BLOCK`, otherwise you'll see [keybase/kbpgp#198](https://github.com/keybase/kbpgp/issues/198)

Interesting, I (along with some help) actually tested the GPG key in a Grafana deployment using Juju with [this bundle file](https://pastebin.ubuntu.com/p/ZhPc5ZFGHT/). Could this maybe be specific to PGP or keybase?

Could you or @morphis try this and confirm?

ajanon (approved):
LGTM

PR #301, opened 2023-09-13
Rearrange addon topics

keirthana:
For https://warthogs.atlassian.net/browse/AC-1734


exp/addons.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+To create an addon, you must provide the Anbox Management Client (AMC) with either of the following:
+* The addon directory
+* A tarball containing the required addon file structure
+* A zip archive containing the required addon file structure
+
+## Best practices
+
+Addons must be created with careful consideration to not affect performance adversely. A good addon is light-weight and are targeted for the necessary applications.
```
ajanon, line 23:
```suggestion
Addons must be created with careful consideration to not affect performance adversely. A good addon is light-weight and is targeted for the necessary applications. 
```

Also, isn't the second sentence a bit redundant with 'Keep addons light' below?

exp/addons.md
```diff
@@ -0,0 +1,42 @@
...
+## Best practices
+
+Addons must be created with careful consideration to not affect performance adversely. A good addon is light-weight and are targeted for the necessary applications. 
+
+Here are some good practices to consider when creating addons:
+
+### Keep addons light
+Addons are executed synchronously. Any addon that performs long-running operations (for example, downloading large files, installing packages on regular containers or querying unresponsive services) will delay an application from starting.
+
+[note type="information" Status="Tip"]Use the `CONTAINER_TYPE` environment variable to run only the necessary code in your hooks.[/note]
```
ajanon, line 30: I feel like this is true, but a big vague. The `CONTAINER_TYPE` environment variable (soon to be `INSTANCE_TYPE` btw) is mostly useful to run only on a given instance type, and thus, run code only _when_ necessary.
Does that make sense? What do you think?

exp/addons.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Use global addons sparingly
+Addons that are enabled for all applications can be useful, but they can add up quickly. Try to attach addons to individual applications unless you need a global addon and you are sure that it won't slow down containers.
+
+### Clean up your addons
+If your addon needs additional tools and dependencies during its installation, make sure you remove them afterwards. This will make your addon lighter and all applications using it will start faster.
```
ajanon, line 36: Shouldn't this be only for base containers/instances, though?
For regular containers, I think that would just increase preparation time.

ajanon (changes requested):

exp/addons.md
keirthana: Yes, but it is intentional. The lead in sentence after "Best practices" is to set the context in the reader's mind while we elaborate more in the later sections. I don't feel strongly about removing it as well. Let me know what you think.

exp/addons.md
keirthana: I have clarified it based on what you explained.

exp/addons.md
ajanon: I think it's fine if it's intentional, thanks for the answer!

ajanon (approved):
LGTM

howto/addons/migrate.md
```diff
@@ -1,9 +1,8 @@
-Starting with Anbox Cloud 1.12, the following hooks are deprecated and should be replaced:
+Starting with Anbox Cloud 1.12, use the following new hooks instead of the deprecated ones:
 
-- `install` should be replaced with `pre-start`
-- `prepare` should be replaced with `post-start`
-- `restore` should be replaced with `pre-start`
-- `backup` should be replaced with `post-stop`
+* Use `pre-start` instead of `install` and `restore`
```
adglkh, line 3: not native speaker but I feel the following feel more nature 
`install` and `restore` 
-> 
`install` or `restore`

howto/addons/migrate.md
```diff
@@ -1,9 +1,9 @@
-Starting with Anbox Cloud 1.12, the following hooks are deprecated and should be replaced:
+Starting with Anbox Cloud 1.12, use the following new hooks instead of the deprecated ones:
 
-- `install` should be replaced with `pre-start`
-- `prepare` should be replaced with `post-start`
-- `restore` should be replaced with `pre-start`
-- `backup` should be replaced with `post-stop`
+* Use `pre-start` instead of `install`
```
adglkh, line 3: hmm, we should drop this line as the `install` hook is addon version free and it doesn't correlate with `pre-start`.
@morphis could you double-check this as well?

adglkh (commented):
One comment, otherwise LGTM

howto/addons/migrate.md
morphis: What do you mean that `install` doesn't correlate with `pre-start`?

adglkh: hmm, if I thought the `install` hook is a special one unlike `restore`, `backup`, `prepare` ones which are bound to a specific addon version 1. But I was wrong obviously and we did deprecate `install` hook for quite a long time.
Everything is clear to me now.
Thanks for the confirmation

adglkh (approved):
LGTM

PR #306, opened 2023-10-03
Replace container terminology

keirthana:
More changes to:
- AMS configuration
- Getting started with dashboard tutorial
- Upgrade appliance procedure
- Install appliance in GCP
- Develop addon
- Prometheus
- Troubleshooting topics

### DIFF
--- a/ref/ams-configuration.yaml
+++ b/ref/ams-configuration.yaml
@@ -43,14 +43,36 @@ all:
     description: >-
       APT mirror to use within the containers. By default,
       `http://archive.ubuntu.com` (amd64) or `http://ports.ubuntu.com`
+      (arm64) is used. This configuration item is deprecated since 1.20,
+      use `instance.apt_mirror` instead.
+  instance.apt_mirror:
+    type: "string"
+    default: "-"
+    description: >-
+      APT mirror to use within the instances. By default,
+      `http://archive.ubuntu.com` (amd64) or `http://ports.ubuntu.com`
       (arm64) is used.
   container.default_platform:
     type: "string"
     default: "-"
     description: >-
-      The name of the platform that Anbox uses by default to launch
-      containers.
+      The name of the platform that Anbox Cloud uses by default to launch
+      containers. This configuration item is deprecated since 1.20,
+      use `instance.default_platform` instead.
+  instance.default_platform:
+    type: "string"
+    default: "-"
+    description: >-
+      The name of the platform that Anbox Cloud uses by default to launch
+      instances.
   container.features:
+    type: "string"
+    default: "-"
+    description: >-
+      Comma-separated list of features to enable (see list below).
+      This configuration item is deprecated since 1.20, use
+      `instance.features` instead.
+  instance.features:
     type: "string"
     default: "-"
     description: >-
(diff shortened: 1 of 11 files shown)


ajanon (approved):
LGTM

howto/troubleshoot/dashboard-issues.md
```diff
@@ -16,32 +16,32 @@ A session does not start and the session details page displays the following err
...
 
 The **Error Message field** can give you a starting point for identifying the issue. Some reasons for an instance to go into error status could be:
 * Insufficient resources. Refer [Capacity planning](https://discourse.ubuntu.com/t/capacity-planning/28717).
-* Occasionally, access to Ubuntu archives could be a problem when creating an application. As an immediate workaround, you could disable the security update by running `amc config set container.security_updates false` or explicitly set `amc config set container.api_mirror <mirror_address>` to configure an instance to use a different APT mirror. See [AMS configuration](https://discourse.ubuntu.com/t/ams-configuration/20872) for more details.
+* Occasionally, access to Ubuntu archives could be a problem when creating an application. As an immediate workaround, you could disable the security update by running `amc config set instance.security_updates false` or explicitly set `amc config set instance.api_mirror <mirror_address>` to configure an instance to use a different APT mirror. See [AMS configuration](https://discourse.ubuntu.com/t/ams-configuration/20872) for more details.
```
jat-canonical, line 37: Is there a need to add a deprecated marker or something to say that the old still works, like it is done below for the API endpoints ?

jat-canonical (commented):
one small question

howto/troubleshoot/dashboard-issues.md
morphis: It should be marked as deprecated in the `ams-configuration.yaml` file I would say and then affects the generated table to mark it as deprecated.

ref/ams-configuration.yaml
```diff
@@ -43,14 +43,36 @@ all:
     description: >-
       APT mirror to use within the containers. By default,
       `http://archive.ubuntu.com` (amd64) or `http://ports.ubuntu.com`
+      (arm64) is used. This configuration item is deprecated since 1.20,
+      use `instance.apt_mirror` instead.
+  instance.apt_mirror:
```
morphis, line 48: As mentioned above we should add a field to indicate the `container.` key variants of these are deprecated.

morphis (changes requested):

ref/ams-configuration.yaml
keirthana: @morphis Do you mean to say instead of adding as text as I have done, I should add a field that would display a separate column for the deprecated status? Currently, I have added that the `container.xxx` variants are deprecated in their respective descriptions. I could shift them to a separate column if that gives more clarity.

howto/troubleshoot/dashboard-issues.md
keirthana: Updated the `ams-configuration.yaml` to add a field for deprecated

ref/ams-configuration.yaml
keirthana: Updated the `ams-configuration.yaml` to add a field for deprecated

ref/ams-configuration.yaml
```diff
@@ -176,53 +244,61 @@ all:
...
       client in AMS operates: `manual`, `pull`, `push`
+    deprecated: false
   registry.update_interval:
     type: "string"
     default: "`1h`"
     description: >-
       Frequency of
       [Anbox Application Registry](https://discourse.ubuntu.com/t/application-registry/17761)
       updates (for example: 1h, 30m).
+    deprecated: false
```
morphis, line 263: We do not necessarily have to put the `false` value as it it considered the default for a boolean but also not a problem, so lets keep as is.

morphis (approved):
LGTM, thanks!

PR #307, opened 2023-10-05
Replace "AMS instances" and "Anbox Cloud instance"

keirthana:
Anbox Cloud instance(s) - This terminology should be avoided because in certain contexts, it could be a deployment of Anbox Cloud as opposed to an Anbox Cloud container or VM. Based on this decision taken during our discussions, the mentions of Anbox Cloud instance(s) are reworded to avoid ambiguity.

AMS instances - Need a decision on this terminology as these are much more context-based. [This](https://chat.canonical.com/canonical/pl/sf1gdefq5pr7dgjtxujd1sb9uh) is the Mattermost discussion thread.
Update: Reworded based on context as per discussions in SU and Mattermost.


ajanon (approved):
LGTM

morphis (approved):
LGTM

PR #309, opened 2023-10-11
Replace screenshots for tutorial

keirthana:
I have replaced the screenshots for the dashboard tutorial. I have removed a few screenshots because they seem too redundant and we do not have to give screenshots for every step. As the dashboard gets more familiar and has more help elements inside it, the eventual goal is to not have screenshots that require more maintenance.

All "instance type" mentions will be verified and adjusted as part of https://warthogs.atlassian.net/browse/AC-2014 and hence is not covered in this PR.


ajanon (approved):
LGTM

adglkh (approved):
LGTM

PR #316, opened 2023-11-09
Pull grafana prometheus

keirthana:
Pulling out all content related to monitoring including Grafana, Prometheus, Nagios, Telegraf


ref/network-ports.md
```diff
@@ -10,7 +10,6 @@ For the Anbox Cloud Appliance, ports are exposed only for accessing the Anbox Cl
 | Service               | Port(s)     | Protocol  | Exposed externally | Required                     | Description                           |
 |-----------------------|-------------|-----------|--------------------|------------------------------|---------------------------------------|
 | AMS                   | 8444        | TCP       | no                 | yes                          | HTTPS API                             |
-| AMS                   | 20002       | TCP       | no                 | no                           | HTTPS Prometheus endpoint             |
```
keirthana, line 13: I removed this because it is offering a prometheus endpoint and we don't talk about it anywhere else. Does that make sense?

ref/prometheus.md
```diff
@@ -1,275 +0,0 @@
-Anbox Cloud gathers various performance metrics that you can access through API endpoints to create a monitoring solution. See [How to monitor Anbox Cloud](https://discourse.ubuntu.com/t/monitor-anbox-cloud/24338) for detailed information.
-
-The reference implementation for Prometheus (see [Example: Collect metrics](https://discourse.ubuntu.com/t/example-collect-metrics/17787)) provides some basic dashboards. You can, however, update them to fit your needs.
-
-The following sections list all metrics returned to Prometheus by the Anbox Cloud services that support Prometheus metrics.
```
keirthana, line 5: @morphis I feel the metrics that AMS provides can still be useful if users are building their own monitoring solution but since these metrics feel tied to prometheus, I removed the entire document. Let me know if we can still list them in a generic way.

ajanon (changes requested):

exp/production.md
```diff
@@ -63,13 +63,7 @@ Load balancing solutions can differ based on your deployment model. You should a
...
 
-You can decide if you want to use Canonical's monitoring system or a customised version. See [How to monitor Anbox Cloud?](https://discourse.ubuntu.com/t/how-to-monitor-anbox-cloud/24338) and [Prometheus metrics](https://discourse.ubuntu.com/t/prometheus-metrics/19521) for a detailed understanding of monitoring and metrics to assess your monitoring needs for the production deployment.
-
-To help you in your assessment, you can refer to the following example implementations:
-* [Example: Collect metrics](https://discourse.ubuntu.com/t/example-collect-metrics/17787)
-* [Example: Monitor status](https://discourse.ubuntu.com/t/example-monitor-status/17788).
-
-Also, see [Performance](https://discourse.ubuntu.com/t/29416) to learn about the factors that could influence the performance of your deployment.
+You can decide if you want to use Canonical's monitoring system or a customised version. Also, see [Performance](https://discourse.ubuntu.com/t/29416) to learn about the factors that could influence the performance of your deployment.
```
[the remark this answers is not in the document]
keirthana: Removed "Also", it's simpler.

howto/monitor/landing.md
```diff
@@ -1,34 +1,7 @@
...
-
-2. Make sure that the monitoring dashboard is exposed. This is the default. If you unexposed it earlier, expose it again by running the following command:
-
-        anbox-cloud-appliance monitor expose
-3. Access the monitoring dashboard at `https://<your-machine-address>/monitor/` and log on with the user account that you created.
-4. Click the search symbol on the left-hand side and go to the **Search dashboards** page.
-5. Select the dashboard that contains the information you're interested in to display it.
-
-If you want to monitor additional metrics and events, you can do so by adding plugins to Telegraf. See [How to configure Telegraf](https://discourse.ubuntu.com/t/how-to-configure-telegraf/30365) for more information.
+There are plans in the Anbox Cloud roadmap to provide better monitoring abilities by using the [Canonical Observability Stack](https://charmhub.io/topics/canonical-observability-stack).
```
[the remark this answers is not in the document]
keirthana: It is a bit tricky. I worded it this way because this is just a reassurance because we are pulling out the ability to monitor totally (although some of the stuff like telegraf and nagios still work but they require additional processing which is tied to Grafana or Prometheus. So we are just pulling out everything from the documentation. But that kind of means that there is no supported monitoring solution for Anbox Cloud in the 1.20 release. So this statement it to reassure the users that we are working on it. I will use your rewording once I add the corresponding version when we are targeting to release integration with COS. Hope that is okay?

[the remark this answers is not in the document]
ajanon: I'm not sure if my version is better than yours, I just wanted to point out the perceived writing style difference.
Thanks for the explanation!

ajanon (approved):
LGTM

ref/prometheus.md
adglkh: We need to keep this page for the sake of people who is going to build their own monitoring stack as you mentioned above.

keirthana: Reverted and brought back content related to prometheus metrics offered. Thanks!

ref/network-ports.md
keirthana: Reverted the change as clarified that we offer the endpoint even if we don't offer a monitoring solution so that customers can build their custom monitoring solution.

morphis (approved):
LGTM

exp/production.md
```diff
@@ -63,13 +63,7 @@ Load balancing solutions can differ based on your deployment model. You should a
...
-
-To help you in your assessment, you can refer to the following example implementations:
-* [Example: Collect metrics](https://discourse.ubuntu.com/t/example-collect-metrics/17787)
-* [Example: Monitor status](https://discourse.ubuntu.com/t/example-monitor-status/17788).
-
-Also, see [Performance](https://discourse.ubuntu.com/t/29416) to learn about the factors that could influence the performance of your deployment.
+You can decide if you want to use Canonical's monitoring system or a customised version. See [Prometheus metrics](https://discourse.ubuntu.com/t/prometheus-metrics/19521) for a detailed understanding of available metrics to assess your monitoring needs for the production deployment. See [Performance](https://discourse.ubuntu.com/t/29416) to learn about the factors that could influence the performance of your deployment.
```
adglkh, line 66: Is the first sentence valid after we dropped the monitoring stack?

adglkh (commented):
One question, otherwise LGTM

exp/production.md
keirthana: Removing it for better clarity. We can add it back after we offer COS integration.

adglkh (approved):
LGTM

PR #318, opened 2023-11-23
fix(ref): Wifi support is enabled by default, update ff doc

ajanon:
The Wifi support feature flag was outdated, as wifi support is enabled by default in Anbox and it can be disabled optionnally, not the other way around.


ref/ams-configuration.md
```diff
@@ -96,7 +96,7 @@ For the feature to be considered, applications must be manually updated, because
 
 #### WiFi
 
-WiFi support can be optionally enabled with the `enable_wifi` feature flag. Anbox will then set up a virtual WiFi device, which sits on top of an Ethernet connection and simulates a real WiFi connection.
+By default, Anbox sets up a virtual WiFi device, which sits on top of an Ethernet connection and simulates a real WiFi connection. This Wifi support can be optionnally disabled with the `disable_wifi` feature flag.
```
morphis, line 99: But only when the `virt_wifi` kernel module is loaded. Otherwise we will fail and silently ignore from what I remember. Maybe we should add this as a explanation/hint.

morphis (changes requested):

ref/ams-configuration.md
ajanon: I will take a closer look and follow-up on this.
The current text is definitely wrong, though.

ajanon: I don't see anything which would test for `virt_wifi` in the code. It is bundled with `anbox-modules` and should be loaded automatically. If `virt_wifi` from `anbox_modules` is not loaded, it probably also means that the `anbox_sync` module is not loaded, and we will have additional problems down the line.

ajanon: I'm not sure if it's necessary to mention anything more here, I think we can reasonably expect virt_wifi to be loaded at that point, unless we have some customers explicitly disabling it.

morphis (approved):
LGTM

keirthana:
Thanks for the fix, this merged and published.

PR #320, opened 2023-12-01
Update documentation for 1.20.1

keirthana:
<s>The component versions need to be updated once the images are built.</s>
Update: Component versions are updated based on [this build](https://next.ci.anbox-cloud.io/view/1.20/job/deployment-scripts-stable-1.20-publish/5/console). Bundle versions may need updates once they are promoted to stable.

### DIFF
--- a/ref/roadmap.md
+++ b/ref/roadmap.md
@@ -12,4 +12,4 @@ Anbox Cloud follows a defined release cycle with frequent minor and patch releas
 | Release | Target Date | What to expect| 
 |---------|-------------|---------------|
 | 1.20.2 | January 17 2024  | * Android security updates for January 2023<br/>* Bug fixes |
-| 1.21.0 | February 14, 2024 | * Production support for Vulkan on NVIDIA GPUs<br/>* Add VHAL support<br/>* Support for IAM role based authentication in AAR<br/>* TLS support for `anbox-cloud-nfs-operator` charm.<br/>* Android security updates for February 2023<br/>* Bug fixes |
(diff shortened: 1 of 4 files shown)


ref/roadmap.md
```diff
@@ -11,5 +11,5 @@ Anbox Cloud follows a defined release cycle with frequent minor and patch releas
 ## Roadmap
 | Release | Target Date | What to expect| 
 |---------|-------------|---------------|
-| 1.20.1 | December 13 2023 | * Android security updates for December 2023<br/>* Bug fixes |
 | 1.20.2 | January 17 2024  | * Android security updates for January 2023<br/>* Bug fixes |
+| 1.21.0 | February 14, 2024 | * Production support for Vulkan on NVIDIA GPUs<br/>* Add VHAL support<br/>* Support for IAM role based authentication in AAR<br/>* TLS support for `anbox-cloud-nfs-operator` charm.<br/>* Android security updates for February 2023<br/>* Bug fixes |
```
ajanon, line 15: Actually, what we really want is automotive support, VHAL is just a part of that.
I think we should mention something like:
> Add Automotive-enabled Anbox Cloud images

Or something along those lines.

(Also, I'm not sure which terminology we are now using for LXD images)

WDYT?

ajanon (changes requested):

ref/roadmap.md
keirthana: Sounds good, the terms are synonymous in our docs but I prefer to use Anbox Cloud images mostly unless there is reason otherwise, in the sense LXD is the technology we use for managing our instances.

ajanon (approved):
LGTM

PR #324, opened 2024-01-16
Invert GL async swap feature

adglkh:
We have GL Async object enabled from the very beginning after the
bring up work for the emugl decoder is complete, however the consequence
of enabling the GL async swap feature is that a few games/apps got
ANRs due to the fact GL call hangs at the host driver caused by this
enabled feature when swapping buffers.

For a better user experience, since 1.21.0 release, this feature
is disabled for Anbox device by default. And we inverted the feature
flag and environment variable accordingly to honor the behavior change.

### DIFF
--- a/ref/ams-configuration.tmpl.md
+++ b/ref/ams-configuration.tmpl.md
@@ -90,9 +90,9 @@ Once set, this feature flag will be considered by all newly launched instances.
 
 #### GL Async Swap Support
 
-*since 1.19.0*
+*since 1.21.0*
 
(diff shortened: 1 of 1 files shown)


adglkh:
@keirthana This is a feature being part of 1.21.0 release. So let's not publish it until 1.21.0 is released.

keirthana:
> @keirthana This is a feature being part of 1.21.0 release. So let's not publish it until 1.21.0 is released.

@adglkh Sure, but can you raise these changes against the https://github.com/canonical/anbox-cloud-docs repo? This discourse-docs repo is archived and no longer the latest repo.

